PR #142, opened 2025-10-19
Fix status comparison case mismatch in processor.py

yrangana:
## Description

Fixed a bug where document status comparisons were using uppercase `"PROCESSED"` string instead of the `DocStatus.PROCESSED` enum, causing `text_processed` to always return `False`.

## Related Issues

This fixes a bug where the status check never matches because:
- LightRAG stores status as lowercase `"processed"` (from DocStatus enum)
- RAGAnything was checking against uppercase `"PROCESSED"` string
- Comparison: `"processed" == "PROCESSED"` always returns `False`

## Changes Made

Replaced hardcoded `"PROCESSED"` strings with `DocStatus.PROCESSED` enum constant in 4 locations:
- Line 481: `process_multimodal_content()` - first status check
- Line 486: `process_multimodal_content()` - second status check  
- Line 1355: `is_document_fully_processed()` - status comparison
- Line 1387: `get_document_processing_status()` - status comparison

Also updated related comments for consistency.

## Benefits

- Fixes the case mismatch bug
- Makes code type-safe (DocStatus enum already imported)
- Removes magic strings
- Future-proof if enum values change

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (if necessary)

## Additional Notes

The `DocStatus` enum was already imported at line 14, so this is just using the existing constant instead of hardcoded strings. Compatible with both LightRAG v1.4.9.2 and v1.4.9.3.

### DIFF
--- a/raganything/processor.py
+++ b/raganything/processor.py
@@ -460,7 +460,7 @@ class ProcessorMixin:
             self.logger.debug("No multimodal content to process")
             return
 
-        # Check multimodal processing status - handle LightRAG's early "PROCESSED" marking
+        # Check multimodal processing status - handle LightRAG's early DocStatus.PROCESSED marking
         try:
             existing_doc_status = await self.lightrag.doc_status.get_by_id(doc_id)
             if existing_doc_status:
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks!

PR #166, opened 2025-11-24
Feat query add system prompt

### LINKED ISSUE #153: [Question]: Passing system_prompt in RAG queries
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [x] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

Hi team,

I noticed that in **LightRAG**, the _aquery_  method accepts a _system_prompt_ parameter:
lightrag/lightrag.py
```
    async def aquery(
        self,
        query: str,
        param: QueryParam = QueryParam(),
        system_prompt: str | None = None,
    ) -> str | AsyncIterator[str]:
        .....
        result = await self.aquery_llm(query, param, system_prompt)  <- def has the **system_prompt** param
```
This allows us to control the LLM’s behavior per query, which is useful for overriding the default system prompt dynamically.

However, in RAG-Anything, the current aquery method does not expose _system_prompt_ directly:
raganything/query.py
```
async def aquery(self, query: str, mode: str = "mix", **kwargs) -> str:
    ...
    result = await self.lightrag.aquery(query, param=query_param)  <- Missing **system_prompt** param
```
This means that all queries use the default system prompt defined in LightRAG or session-level configuration.

### Additional Context

Could we expose a system_prompt parameter in RAG-Anything’s aquery method as well? This would allow developers to:
Dynamically adjust the AI behavior per query.
Override the default system prompt without changing the session-level configuration.
Leverage LightRAG’s existing system_prompt support more effectively.

Example of proposed signature:
`async def aquery(self, query: str, mode: str = "mix", system_prompt: str | None = None, **kwargs) -> str`

Then pass system_prompt down to LightRAG:
`result = await self.lightrag.aquery(query, param=query_param, system_prompt=system_prompt)`

EightyOliveira:
## Description
- Removed unused `exclude` sections from the pre-commit configuration file  
- Added a `system_prompt` parameter to the `aquery` function

## Related Issues

Resolves #153

## Changes Made

1. Removed all unnecessary `exclude` entries from `.pre-commit-config.yml`  
2. Added a third parameter `system_prompt` to the `aquery` function. When the query is text-only (non-multimodal), it passes this parameter directly to LightRAG’s `query` method.  
3. Added `system_prompt` support to `aquery_vlm_enhanced`, and included it when constructing system messages in `_build_vlm_messages_with_images`.

## Checklist

- [ ] Changes tested locally  
- [ ] Code reviewed  
- [ ] Unit tests added (if applicable)

## Additional Notes

- The linting tools in use are **Ruff**  or **Black** ??? with Ruff configured as the linter in pre-commit.
- Why must `load_dotenv()` be called before the imports? Won’t this trigger an **E402** ("module level import not at top of file") error from Ruff?
- Our `pyproject.toml` does **not** disable or ignore E402—so how is this code passing the linter?

Example snippet that raises the concern:
```python
# Add project root directory to Python path
sys.path.insert(0, str(Path(__file__).parent.parent))

# Load environment variables from .env file BEFORE importing LightRAG
# This is critical for TIKTOKEN_CACHE_DIR to work properly in offline environments
# The OS environment variables take precedence over the .env file
load_dotenv(dotenv_path=".env", override=False)

from lightrag import LightRAG
from lightrag.utils import logger
```

### DIFF
--- a/raganything/query.py
+++ b/raganything/query.py
@@ -97,7 +97,9 @@ class QueryMixin:
 
         return f"multimodal_query:{cache_hash}"
 
-    async def aquery(self, query: str, mode: str = "mix", system_prompt: str | None = None,**kwargs) -> str:
+    async def aquery(
+        self, query: str, mode: str = "mix", system_prompt: str | None = None, **kwargs
+    ) -> str:
         """
         Pure text query - directly calls LightRAG's query functionality
 
@@ -134,7 +136,9 @@ class QueryMixin:
             and hasattr(self, "vision_model_func")
             and self.vision_model_func
         ):
-            return await self.aquery_vlm_enhanced(query, mode=mode, system_prompt=system_prompt,**kwargs)
+            return await self.aquery_vlm_enhanced(
+                query, mode=mode, system_prompt=system_prompt, **kwargs
+            )
         elif vlm_enhanced and (
             not hasattr(self, "vision_model_func") or not self.vision_model_func
         ):
@@ -149,7 +153,9 @@ class QueryMixin:
         self.logger.info(f"Query mode: {mode}")
 
         # Call LightRAG's query method
-        result = await self.lightrag.aquery(query, param=query_param,system_prompt=system_prompt)
+        result = await self.lightrag.aquery(
+            query, param=query_param, system_prompt=system_prompt
+        )
 
         self.logger.info("Text query completed")
         return result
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks for your contribution! However, there are a few lint errors that need to be fixed. Could you please run `pre-commit run --all-files` to resolve them?

LarFii:
> Our pyproject.toml does not disable or ignore E402—so how is this code passing the linter?

Thanks for your contribution again! We ignore E402 in our pre-commit config.

```
- id: ruff
  args: [--fix, --ignore=E402]
```

PR #208, opened 2026-02-20
fix: handle missing event loop in close() during interpreter shutdown

### LINKED ISSUE #135: [Bug]: Warning: Failed to finalize RAGAnything storages: There is no current event loop in thread 'MainThread'.

### Describe the bug

Just used the demo script in the README, got that warning

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

haosenwang1018:
Closes #135

When the `atexit` handler calls `close()`, the default event loop policy may already be torn down during interpreter shutdown, causing `asyncio.run()` to raise `RuntimeError: There is no current event loop in thread 'MainThread'`.

This fix adds a fallback: if `asyncio.run()` fails, we manually create a new event loop with `asyncio.new_event_loop()` and run the finalization there. This ensures storages are properly finalized regardless of when the cleanup runs.

### DIFF
--- a/raganything/raganything.py
+++ b/raganything/raganything.py
@@ -141,12 +141,22 @@ class RAGAnything(QueryMixin, ProcessorMixin, BatchMixin):
             # Check if there's a running event loop using get_running_loop()
             # This is the proper way in Python 3.10+ to avoid DeprecationWarning
             try:
-                asyncio.get_running_loop()
+                loop = asyncio.get_running_loop()
                 # If we're in an async context, schedule cleanup
                 asyncio.create_task(self.finalize_storages())
             except RuntimeError:
-                # No running event loop, run cleanup synchronously
-                asyncio.run(self.finalize_storages())
+                # No running event loop — create a new one to run cleanup.
+                # asyncio.run() can fail during interpreter shutdown when the
+                # default event-loop policy has already been torn down, so we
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks for the fix.

I reviewed the change and found two potential issues:

1. **Blocking: `RuntimeError` handling is too broad in the fallback path**  
   In `raganything/raganything.py`, the new `except RuntimeError:` around `asyncio.run(self.finalize_storages())` may also catch `RuntimeError` raised *inside* `finalize_storages()` itself (not only event-loop-policy teardown errors).  
   That can:
   - mask the original failure reason,
   - trigger a second finalize attempt unexpectedly,
   - make debugging harder.

2. **Non-blocking: manual loop should be set as current loop during fallback**  
   In the `new_event_loop()` branch, it is safer to call `asyncio.set_event_loop(loop)` before `run_until_complete(...)`, then clear it with `asyncio.set_event_loop(None)` in `finally`.  
   Some downstream code may still call `asyncio.get_event_loop()` during finalize; without setting the current loop, fallback can still fail in those cases.

haosenwang1018:
Closing to reduce stale review backlog.

PR #209, opened 2026-02-20
fix: strip thinking tags before all JSON parse strategies

### LINKED ISSUE #159: [Bug]: Analyzed data of tables not complete

### Describe the bug

When running table or chunk analysis using a reasoning-enabled model (such as qwen2.5-think, deepseek-r1, or similar), the system sometimes stores only the model’s internal “thinking” text (<think> ... </think>) instead of the real analysis result.

```
{
  "content": "<think>\nOkay, let's see. The user wants me to analyze this table and provide a JSON response with sp..."
}
```

The final structured output or description (which should follow the reasoning) is missing completely.

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

haosenwang1018:
Closes #159

When using reasoning models (qwen2.5-think, deepseek-r1, etc.), the `<think>...</think>` tags were only stripped in `_extract_all_json_candidates()`, meaning the regex fallback strategy (`_extract_fields_with_regex`) still operated on the raw response including thinking content. This could cause it to extract content from the thinking section rather than the actual analysis.

This fix moves the think-tag stripping to the top of `_robust_json_parse()` so all downstream strategies work with clean model output.

### DIFF
--- a/raganything/modalprocessors.py
+++ b/raganything/modalprocessors.py
@@ -546,6 +546,18 @@ class BaseModalProcessor:
 
     def _robust_json_parse(self, response: str) -> dict:
         """Robust JSON parsing with multiple fallback strategies"""
+        import re
+
+        # Pre-strip thinking/reasoning tags so all strategies (including the
+        # regex fallback) operate on the actual model output, not internal
+        # chain-of-thought text. Fixes issues with reasoning models like
+        # qwen2.5-think and deepseek-r1 (#159).
+        response = re.sub(
+            r"<think>.*?</think>", "", response, flags=re.DOTALL | re.IGNORECASE
+        )
+        response = re.sub(
+            r"<thinking>.*?</thinking>", "", response, flags=re.DOTALL | re.IGNORECASE
+        )
 
         # Strategy 1: Try direct parsing first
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks for the fix.

I do see one potential side effect to consider:

- **Potential data loss from global tag stripping**  
  The current implementation removes `<think>...</think>` / `<thinking>...</thinking>` across the entire response before all parsing strategies.  
  If those tags appear as legitimate literal content inside the actual payload (e.g., in `detailed_description`), that content would be removed unintentionally.

Also a minor maintainability point:

- **Duplicated cleanup logic**  
  The same think-tag stripping still exists in `_extract_all_json_candidates()`, so cleanup now happens in two places. It may be better to centralize this in one function to avoid drift.

### Suggested refinement
Instead of global removal, strip only **leading reasoning blocks** (prefix-only), which still fixes the fallback issue while avoiding accidental mutation of valid body content.

haosenwang1018:
Closing per review backlog policy; can resubmit in smaller chunks.

PR #210, opened 2026-02-21
feat: support environment variables in parsers

wkpark:
support environment variables in parsers

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

you can set environment variables for Mineru and Docling.

for example, you can set `CUDA_VISIBLE_DEVICES=1` to make the parser run on a specific GPU. This prevents the parser from slowing down your main AI model on the first GPU.

```py
# Use the second GPU (ID: 1) for this parsing task
    file_path="manual.pdf",
    env={"CUDA_VISIBLE_DEVICES": "1"}
)
```

### DIFF
--- a/raganything/parser.py
+++ b/raganything/parser.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 
 import os
+import hashlib
 import json
 import argparse
 import base64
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks for the contribution.

One potential regression to consider:
- In `MineruParser._run_mineru_command`, the new `**kwargs` makes unknown arguments silently ignored (except `env`) instead of failing fast. Previously, unexpected arguments would raise a `TypeError`, which was easier to catch during integration.
- This could lead to “runs successfully but config not applied” scenarios.

Suggestion:
1. Validate and explicitly reject unsupported `kwargs` (or at least log a warning).
2. Add light validation for `env` (e.g., ensure mapping of string keys/values).
3. Add tests for:
   - `env` propagation to subprocess
   - behavior when unknown kwargs are passed

LarFii:
@codex review

Issue #172, opened 2025-12-05
[Bug]:The process got stuck at rag.process_document_complete due to network issues.


xtanitfy:
### Do you need to file an issue?

- [ ] I have searched the existing issues and this bug is not already filed.
- [ ] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

Since it's deployed in a local area network (LAN), can it be switched to offline mode?

### LightRAG Config Used

# Paste your config here

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

xtanitfy:
Since it's deployed in a local area network (LAN), can it be switched to offline mode?

xtanitfy:
我准备下载 https://www.modelscope.cn/models/OpenDataLab/MinerU2.5-2509-1.2B，先使用这个做测试，但是还不知道怎么把这个离线的加入框架

PR #216, opened 2026-03-02
feat: add retry and circuit breaker utilities for LLM calls (mitigates #172)

Jah-yee:
### Summary
This PR adds a small `raganything.resilience` module with reusable retry and circuit breaker helpers for LLM API calls, so long-running document processing can better tolerate transient network issues.

### Motivation
As discussed in #172, `process_document_complete` and similar flows can get stuck when LLM calls intermittently fail: there is no retry/backoff strategy and no circuit breaker to prevent cascading failures. A focused resilience layer makes it easier to harden these call sites without pulling in extra dependencies.

### Changes
- **`raganything/resilience.py`**
  - `@retry` decorator for synchronous functions: exponential backoff with optional jitter, detection of common transient exceptions (`httpx`, OpenAI clients, generic network errors), configurable attempts/delays, optional `on_retry` callback.
  - `@async_retry` decorator with the same semantics for async functions.
  - `CircuitBreaker` class: tracks consecutive failures, opens at a configurable threshold, uses half-open trials to recover automatically when the upstream stabilizes. In-memory and dependency-free.
- **`tests/test_resilience.py`**
  - Tests for sync/async retry behavior, jitter/backoff ranges, `on_retry` callback invocation, circuit breaker state transitions (closed → open → half-open → closed), and error handling.

### Testing
- Ran `pytest` locally including `tests/test_resilience.py`; all tests passed.
- Manually simulated intermittent failures to confirm retries, backoff, and breaker recovery behave as expected.

Thanks for your work on RAG-Anything—if you’d like different defaults or naming for these helpers, I’m happy to revise the PR to match your preferences.

### DIFF
--- a/raganything/__init__.py
+++ b/raganything/__init__.py
@@ -108,6 +108,7 @@ if "set_prompt_language" in globals():
         ]
     )
 
+
 def get_version() -> str:
     """Return the RAG-Anything version string."""
     return __version__
--- a/raganything/resilience.py
+++ b/raganything/resilience.py
@@ -1,0 +1,316 @@
+"""
+Retry and resilience utilities for RAGAnything.
+
+Provides decorators and helpers for handling transient failures in LLM API
+calls, embedding requests, and other network-dependent operations.
+
+Addresses GitHub issue #172 — process_document_complete getting stuck due to
+intermittent network errors.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import functools
+import logging
+import time
+from typing import (
+    Any,
+    Callable,
+    Optional,
+    Sequence,
+    Type,
+    TypeVar,
+    Union,
+)
+
+logger = logging.getLogger(__name__)
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+# Default transient exceptions that are safe to retry
+_DEFAULT_RETRYABLE: tuple[Type[BaseException], ...] = (
+    ConnectionError,
+    TimeoutError,
+    OSError,
+)
+
+try:
+    import httpx
+
+    _DEFAULT_RETRYABLE = _DEFAULT_RETRYABLE + (
+        httpx.ConnectError,
+        httpx.ReadTimeout,
+        httpx.WriteTimeout,
+        httpx.PoolTimeout,
+    )
+except ImportError:
+    pass
+
+try:
+    import openai
+
+    _DEFAULT_RETRYABLE = _DEFAULT_RETRYABLE + (
+        openai.APIConnectionError,
+        openai.APITimeoutError,
+        openai.RateLimitError,
+        openai.InternalServerError,
+    )
+except ImportError:
+    pass
+
+
+def retry(
+    max_attempts: int = 3,
+    base_delay: float = 1.0,
+    max_delay: float = 60.0,
+    exponential_base: float = 2.0,
+    jitter: bool = True,
+    retryable_exceptions: Optional[Sequence[Type[BaseException]]] = None,
+    on_retry: Optional[Callable[[BaseException, int, float], None]] = None,
+) -> Callable[[F], F]:
+    """Decorator that retries a **synchronous** function on transient failures.
+
+    Uses exponential backoff with optional jitter to avoid thundering-herd
+    problems when multiple workers hit rate limits simultaneously.
+
+    Args:
+        max_attempts: Total number of attempts (including the first call).
+        base_delay: Initial delay in seconds between retries.
+        max_delay: Upper-bound on the delay between retries.
+        exponential_base: Multiplier applied to the delay after each retry.
+        jitter: If ``True``, adds random jitter (0–50 % of computed delay).
+        retryable_exceptions: Exception types that trigger a retry.
+            Defaults to common network / API transient errors.
+        on_retry: Optional callback ``(exception, attempt, delay)`` invoked
+            before each retry sleep.
+
+    Returns:
+        The decorated function, with retry behaviour.
+
+    Example::
+
+        @retry(max_attempts=5, base_delay=2.0)
+        def call_llm(prompt: str) -> str:
+            return openai.ChatCompletion.create(...)
+    """
+    if retryable_exceptions is None:
+        retryable_exceptions = _DEFAULT_RETRYABLE
+
+    def decorator(func: F) -> F:
+        @functools.wraps(func)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            last_exception: BaseException | None = None
+            for attempt in range(1, max_attempts + 1):
+                try:
+                    return func(*args, **kwargs)
+                except tuple(retryable_exceptions) as exc:
+                    last_exception = exc
+                    if attempt == max_attempts:
+                        logger.error(
+                            "%s failed after %d attempts: %s",
+                            func.__qualname__,
+                            max_attempts,
+                            exc,
+                        )
+                        raise
+                    delay = min(
+                        base_delay * (exponential_base ** (attempt - 1)),
+                        max_delay,
+                    )
+                    if jitter:
+                        import random
+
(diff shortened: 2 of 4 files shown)


LarFii:
Thanks for the resilience utilities work. I did a **Codex-assisted review** and found two blocking behavior issues to fix before merge:

1. **Half-open state does not enforce a single trial call (High)**
   - In `CircuitBreaker`, rejection only happens when `state == "open"`. After timeout, `state` becomes `"half-open"`, but there is no gate to limit admission.
   - Result: multiple concurrent calls can pass through during half-open, which conflicts with the docstring (“allows one trial call through”) and reduces protection during recovery.
   - **Suggested fix:** add half-open single-flight control (e.g., lock + `trial_in_flight` flag, or semaphore=1). While one probe is in flight, additional calls should be rejected with `CircuitBreakerOpen`.
   - Please add a concurrency test to verify only one half-open trial is admitted.

2. **Invalid `max_attempts` produces misleading runtime failure (Medium)**
   - In both `retry` and `async_retry`, if `max_attempts <= 0`, the retry loop is skipped and execution reaches `raise last_exception` with `last_exception is None`, causing `TypeError: exceptions must derive from BaseException`.
   - **Suggested fix:** validate parameters at decorator creation (`max_attempts >= 1`, non-negative delays, sensible bounds) and raise clear `ValueError` for invalid configs.
   - Please add tests for invalid parameter values.

Jah-yee:
Thanks a lot for the detailed Codex-assisted review on the resilience utilities — both issues you highlighted were very helpful.

**Half-open single-flight behaviour**
-  now enforces a true single trial call when transitioning to half-open:
  - state and counters are guarded by a , with a  flag.
  - once the breaker is half-open, the first caller is admitted and marked as the in-flight trial, and any concurrent calls while that trial is running are rejected with .
  - on success or failure, the trial flag and state are reset appropriately ( on success,  on failure).
- I added a concurrency test () that spawns several threads in the half-open window and asserts exactly one successful trial execution and that the remaining callers receive .

**Retry parameter validation**
- Both  and  now validate their configuration at decorator creation time:
  -  is required.
  -  and  must be non-negative.
  -  must be strictly greater than 0.
- This prevents the  path and instead raises a clear  when the configuration is invalid.
-  has been extended with tests covering invalid  and delay values for both sync and async decorators.

If you would like different defaults or error types for misconfiguration, I am happy to tweak them to match your preferred style.

Jah-yee:
Thanks a lot for the detailed Codex-assisted review on the resilience utilities — both issues were very helpful.

- Half-open single-flight behaviour: CircuitBreaker now enforces a true single trial call when transitioning to half-open, using a threading.Lock and a _trial_in_flight flag. The first caller after timeout is admitted as the trial; concurrent callers while the trial is in progress are rejected with CircuitBreakerOpen. On success or failure the state and flag are reset appropriately. A new test (test_half_open_allows_single_trial_call) covers this behaviour under concurrency.

- Retry parameter validation: both retry and async_retry now validate their configuration at decorator creation time (max_attempts >= 1, non-negative base_delay/max_delay, exponential_base > 0) and raise a clear ValueError for invalid configs. tests/test_resilience.py has been extended with tests for invalid max_attempts and delay values for both sync and async decorators.

If you prefer different defaults or error types for misconfiguration, I’m happy to tweak them.

LarFii:
I found two blocking issues in the current resilience defaults.

First, the circuit breaker records every `Exception` as an upstream failure. In both the sync and async wrappers, any application bug such as `TypeError`, `ValueError`, or `KeyError` increments the breaker and can eventually open it. That means local programming errors get misclassified as service instability, and later healthy calls are rejected for the wrong reason. I think the breaker needs the same kind of exception filtering as `retry`, with defaults limited to clearly transient upstream failures.

Second, `_DEFAULT_RETRYABLE` is too broad because it includes `OSError`. Since many permanent local errors inherit from `OSError` (`FileNotFoundError`, `PermissionError`, etc.), this default will retry failures that should surface immediately instead of being delayed by backoff. For a resilience helper intended for LLM/network calls, the default retry set should be narrowed to actual transient network/API exceptions.

Because of those two behaviors, I don’t think this is safe to merge yet.

Jah-yee:
Thanks a lot for the careful second review on the resilience helpers — both points about over-broad failure classification were spot on.

I pushed an update that addresses the two blocking behaviours you raised:

**1) Circuit breaker now filters failures like retry does**

- `CircuitBreaker` accepts an optional `failure_exceptions` parameter and by default uses the same narrow `_DEFAULT_RETRYABLE` set as the retry decorators.
- The sync/async wrappers now only call `record_failure()` when the raised exception is in this configured set; application bugs such as `TypeError`, `ValueError`, `KeyError`, etc. no longer contribute to opening the breaker.
- For non-transient exceptions, the breaker leaves its state unchanged but makes sure to clear the half-open trial gate so a local bug can’t strand the breaker in a permanent `half-open, trial_in_flight=True` limbo.
- New test `test_application_error_does_not_trip_breaker` verifies that a `TypeError` does not move the breaker out of `closed` or increment the failure count.

**2) Default retry set is narrowed and excludes broad `OSError`**

- `_DEFAULT_RETRYABLE` has been tightened to only include connection/timeouts and the httpx/OpenAI transient API errors; the built-in `OSError` base class has been removed from the default set so permanent local issues like `FileNotFoundError` and `PermissionError` aren’t silently retried.
- Call sites that really want to treat local OS errors as retryable can still pass an explicit `retryable_exceptions` list, but the out-of-the-box behaviour is now aligned with LLM/network-centric failure modes.
- A new test `test_default_retry_does_not_retry_oserror_subclasses` asserts that a default-decorated function raising `FileNotFoundError` only executes once and is not retried.

If you’d like the default failure set for the breaker to be even narrower/different from `_DEFAULT_RETRYABLE` (e.g. a separate `_DEFAULT_BREAKER_FAILURES` tuned specifically for LLM calls), I’m happy to split that out as well.

Jah-yee:
@LarFii Thanks again for the detailed review and for flagging the pre-commit issues.

Quick update on the latest changes:

- I merged the latest `main` into `feat-resilience-llm-calls` so the branch now includes the new parser/registry exports, callback/prompt helpers, and the new tests (including `tests/test_core_modules.py`).
- I then applied the exact diff that `pre-commit run --all-files` was generating in CI:
  - Added the blank line before `get_version()` in `raganything/__init__.py` so the export block and the function definition are separated the way `ruff-format` expects.
  - Removed the trailing-whitespace-only line in `tests/test_core_modules.py` (between the `filter_supported_files` call and the `supported_names` assignment), matching the `trailing-whitespace` hook output.
- On top of the earlier resilience updates (narrowed `_DEFAULT_RETRYABLE`, circuit breaker failure filtering, and the new tests), the branch should now be in a state where `pre-commit` runs cleanly in the `lint-and-format` job.

If you see any remaining linting or style issues, I’m happy to align further with your preferred conventions.

LarFii:
No worries, I’ll fix all the lint errors together in the final pass.

LarFii:
You changed `CircuitBreaker` so that, by default, it only counts exceptions in `_DEFAULT_RETRYABLE` toward opening the breaker. In the final PR state, that means generic `RuntimeError` no longer increments `_failure_count` or transitions the breaker to `open`.

However, the tests still assume the old behavior. For example, `test_opens_after_threshold`, `test_half_open_after_timeout`, and `test_half_open_allows_single_trial_call` all use `RuntimeError("fail")` and expect the breaker to open / become half-open. That no longer matches the implementation.

I reproduced this against the PR head: after two `RuntimeError` failures, the breaker stays `closed` and `failure_count` remains `0`.

So there is a mismatch between the implementation and the test/expected semantics. If narrowing the default failure set is intentional, these tests should use a retryable exception such as `ConnectionError` or explicitly pass `failure_exceptions=[RuntimeError]`. If it is not intentional, the implementation should be adjusted.

PR #220, opened 2026-03-02
feat: add multilingual prompt template support (closes #85)

### LINKED ISSUE #85: [Feature Request]: promt "Use {language} as output language" support

While we can change the default text processing language, we should also be able to set the language for processing images, tables, and equations via an environment variable.

### Additional Context

_No response_

Jah-yee:
### Summary
This PR adds first-class support for multilingual prompt templates, including a complete Chinese (`zh`) set and a small runtime manager for switching prompt languages with lazy loading and safe fallback to English.

### Motivation
Right now, all prompt templates are hard-coded in English, which can reduce RAG quality and user experience for non-English users (#85). Different languages often need different phrasing for instructions and system messages to work well with LLMs. Multilingual prompts and a simple runtime selector help RAG-Anything better serve a global user base.

### Changes
- **`raganything/prompts_zh.py`**
  - Full set of Chinese prompt templates (27 keys) mirroring the existing English keys, using natural, idiomatic Chinese phrasing for LLMs.
- **`raganything/prompt_manager.py`**
  - `set_prompt_language(lang)` to switch the active prompt language at runtime (e.g. `"zh"`, `"en"`).
  - `reset_prompts()` to restore the default language and state.
  - `register_prompt_language(lang, prompts)` to register additional languages (e.g. `"ja"`) without changing core code.
  - Lazy loading of language modules so unused languages do not add overhead.
  - Fallback to English when a key is missing in the selected language.
- **`tests/test_prompt_language.py`**
  - Tests for language switching, lazy loading behavior, registering a new language, fallback semantics when keys are missing, and `reset_prompts()`.

### Testing
- Ran `pytest` locally including `tests/test_prompt_language.py`; all tests passed.
- Manually tested switching between `"en"` and `"zh"` to confirm correct localized templates and English fallback where needed.

Thanks for your work on RAG-Anything—if you’d like different language codes, file naming, or prompt phrasing, I’m happy to revise this PR.


LarFii:
I did a **Codex-assisted review** of this PR and found a few issues worth addressing before merge:

1. **Config-based activation is documented but not implemented (P1)**  
   In `prompts_zh.py`, the docstring says users can enable Chinese prompts via `RAGAnythingConfig.prompt_language = "zh"`.  
   However, this PR does not add `prompt_language` to `RAGAnythingConfig`, and I don’t see initialization wiring in `RAGAnything` that calls `set_prompt_language(...)`.  
   So currently this works only via manual runtime API calls, which is a doc/behavior mismatch.

2. **Language code normalization is inconsistent (P2)**  
   `set_prompt_language()` normalizes input with `strip().lower()`, but `register_prompt_language()` stores language codes as-is.  
   Example: registering `"FR"` then switching to `"fr"` fails.  
   Suggestion: normalize codes in `register_prompt_language()` too, and add a test for mixed-case registration.

3. **Global prompt mutation is not concurrency-safe (P2)**  
   `set_prompt_language()` mutates the global `PROMPTS` dict key-by-key.  
   Under concurrent requests, readers can observe partially switched state (mixed-language prompts).  
   Suggestion: make switching atomic (at least lock-protected), or better scope prompt sets per instance/request instead of global mutable state.

Jah-yee:
Thanks a lot for the Codex-assisted review on the multilingual prompts PR.

- Config-based activation: RAGAnythingConfig now has a prompt_language field (backed by PROMPT_LANGUAGE), and RAGAnything.__post_init__ reads it and calls set_prompt_language(...) at initialization time, so the documented RAGAnythingConfig.prompt_language = "zh" flow is now actually wired into the main pipeline.
- Language code normalization: prompt_manager now uses a shared _normalize_language_code() helper in both set_prompt_language() and register_prompt_language(). Codes are normalized with strip().lower() at registration time as well, so registering "FR" and then switching with set_prompt_language("fr") works as expected. tests/test_prompt_language.py now includes a mixed-case registration test.
- Concurrency safety: I introduced a threading.RLock around PROMPTS updates. set_prompt_language() now builds a resolved prompts dict (target language plus English fallback), then swaps PROMPTS and updates _current_language inside the lock; reset_prompts() does the same to restore the English baseline. This avoids partially switched mixed-language states under concurrent access.

If you’d like prompt_language to default to None (opt-in only) or prefer a different env var name, I’m happy to tweak it.

LarFii:
You wired `config.prompt_language` into `RAGAnything.__post_init__()`, but `set_prompt_language()` mutates the process-global `raganything.prompt.PROMPTS` dictionary. That means this is no longer just an instance setting: creating one `RAGAnything` with `prompt_language="zh"` and then another with `"en"` will silently switch the first instance back to English as well. Since both `query.py` and `modalprocessors.py` read from the shared global `PROMPTS` at runtime, different instances/tasks in the same process can overwrite each other’s prompt language. This looks like a real behavioral regression for multi-instance or concurrent usage, so I don’t think `prompt_language` can be exposed as a config field unless prompt selection becomes instance-scoped instead of global.

Jah-yee:
Thanks a lot for calling out the instance-vs-global mismatch here — you’re absolutely right that wiring `config.prompt_language` into `RAGAnything.__post_init__` made this look like an instance-scoped setting while still mutating the process-global `raganything.prompt.PROMPTS`, which is a real regression for multi-instance/concurrent usage.

Given that prompt selection is fundamentally process-global in the current design, I went with the conservative fix rather than a larger refactor:

- **Removed `prompt_language` from `RAGAnythingConfig` and the `__post_init__` wiring**
  - `RAGAnythingConfig` no longer exposes a `prompt_language` field.
  - `RAGAnything.__post_init__` no longer reads config and calls `set_prompt_language(...)`, so creating separate `RAGAnything` instances will not silently override each other’s prompts via a shared global.
- **Clarified that prompt language is a process-global knob**
  - Updated `raganything/prompts_zh.py` docstring to drop the config-based activation and point users to `set_prompt_language("zh")` in `raganything.prompt_manager` as the correct way to switch languages at process level.
  - Updated `raganything/prompt_manager.py` usage docs to explicitly describe `set_prompt_language` as a **process-global** switch.

This keeps the existing multilingual prompt support (including lazy loading, normalization, and the locking to avoid mixed-language states under concurrency), but avoids exposing `prompt_language` as an instance-level config until/unless prompts are refactored to be instance-scoped.

If you’d prefer a follow-up that introduces an explicit `process_prompt_language` env var or a dedicated process-level settings object (so users can still configure the default without touching instances), I’m happy to add that as a separate change.

Jah-yee:
@LarFii Quick follow-up on the linting side for the multilingual prompts PR:

- I pulled in the latest `tests/test_core_modules.py` from `main` and then applied exactly the whitespace fix that `pre-commit` was generating (removing the trailing-whitespace-only line between the `filter_supported_files` call and the `supported_names` assignment).
- I also committed the earlier `ruff-format` changes for `raganything/prompts_zh.py` and the long-assert formatting in `tests/test_prompt_language.py` so that `ruff-format` and `ruff` now run cleanly on the branch.

With those adjustments pushed, `pre-commit run --all-files` should now be a no-op for this branch, and the `lint-and-format` workflow is expected to pass on the next run.

LarFii:
`[P1] Language switches are not actually atomic for concurrent readers`

`set_prompt_language()` / `reset_prompts()` currently mutate the shared global \`PROMPTS\` dict in place via \`PROMPTS.clear()\` followed by \`PROMPTS.update(...)\` ([raganything/prompt_manager.py]). Although the write side is protected by \`_PROMPTS_LOCK\`, existing readers across the codebase still access \`PROMPTS["..."]\` without taking that lock. That means a concurrent request can observe the dict between `clear()` and `update()`, which can raise `KeyError` or produce a partially switched prompt set mid-request. Since this feature is documented as a process-global runtime switch, that makes prompt language changes unsafe in a live multi-threaded service. I’d recommend avoiding in-place mutation of the shared dict here, e.g. by using a copy-on-write approach or by routing all prompt reads/writes through a synchronized accessor.`

PR #228, opened 2026-03-21
fix: preserve full_entities metadata when adding multimodal entities

Exploreunive:
## Summary

This fixes a data-loss bug in Stage 3.5 multimodal entity storage.

When `_store_multimodal_entities_to_full_entities()` updates an existing
`full_entities` entry, it previously rebuilt the record using only:

- `entity_names`
- `count`
- `update_time`

As a result, any existing metadata fields already stored for the same `doc_id`
could be silently overwritten.

## Changes

- preserve existing `full_entities` metadata via merge instead of rebuilding from scratch
- append only new multimodal entity names without duplicating existing ones
- preserve the original entity order instead of converting through a set
- add regression tests covering:
- new entry creation
- preserving existing metadata fields
- deduplicated append behavior without reordering

## Why this matters

Stage 3.5 should enrich existing document entity records with multimodal main
entities, not replace previously stored metadata from the text pipeline.

## Testing

```bash
pytest -q tests/test_full_entities_merge.py tests/test_close_event_loop.py

### DIFF
--- a/raganything/processor.py
+++ b/raganything/processor.py
@@ -1193,31 +1193,32 @@ class ProcessorMixin:
 
             if current_doc_entities is None:
                 # Create new document entry
-                entity_names = list(
+                entity_names = [
                     entity_data["entity_name"]
                     for entity_data in entities_to_store.values()
-                )
+                ]
                 doc_entities_data = {
                     "entity_names": entity_names,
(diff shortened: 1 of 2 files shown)


LarFii:
Thanks for your contribution!

PR #231, opened 2026-03-24
feat(examples): make LLM and vision model names configurable via env vars

zhangzhenfei:
## Summary

Currently, LLM and vision model names (`gpt-4o-mini`, `gpt-4o`) are hardcoded in example scripts. Users who rely on OpenAI-compatible providers (e.g. Azure, local inference servers, third-party proxies) must edit the Python source to change models — while `EMBEDDING_MODEL` and `EMBEDDING_DIM` are already read from environment variables.

This PR brings LLM/vision model configuration in line with the existing embedding pattern.

### Changes

- **`example_knowledge_qa.py`**: read `LLM_MODEL`, `VISION_MODEL`, `EMBEDDING_MODEL`, `EMBEDDING_DIM` from env (defaults preserved: `gpt-4o-mini`, `gpt-4o`, `text-embedding-3-large`, `3072`)
- **`examples/raganything_example.py`**: read `LLM_MODEL`, `VISION_MODEL` from env (same defaults)
- **`env.example`**: document the new `VISION_MODEL` variable alongside `LLM_MODEL`

### Usage after this change

Users can now switch models without touching code:

```dotenv
# .env
LLM_MODEL=kimi-k2.5
VISION_MODEL=kimi-k2.5
EMBEDDING_MODEL=text-embedding-3-large
EMBEDDING_DIM=3072
```

## Test plan

- [ ] Verify defaults are unchanged (existing OpenAI users unaffected)
- [ ] Verify model names are picked up correctly when env vars are set
- [ ] `examples/lmstudio_integration_example.py` already uses this pattern — confirm consistency

### DIFF
--- a/env.example
+++ b/env.example
@@ -112,6 +112,7 @@ MAX_TOKENS=32768
 ### LLM Binding type: openai, ollama, lollms, azure_openai, lmstudio, vllm
 LLM_BINDING=openai
 LLM_MODEL=gpt-4o
+VISION_MODEL=gpt-4o
 LLM_BINDING_HOST=https://api.openai.com/v1
 LLM_BINDING_API_KEY=your_api_key
 ### Optional for Azure
--- a/examples/raganything_example.py
+++ b/examples/raganything_example.py
@@ -116,9 +116,12 @@ async def process_with_rag(
         )
 
         # Define LLM model function
+        llm_model = os.getenv("LLM_MODEL", "gpt-4o-mini")
+        vision_model = os.getenv("VISION_MODEL", "gpt-4o")
+
         def llm_model_func(prompt, system_prompt=None, history_messages=[], **kwargs):
             return openai_complete_if_cache(
-                "gpt-4o-mini",
+                llm_model,
                 prompt,
                 system_prompt=system_prompt,
                 history_messages=history_messages,
@@ -139,7 +142,7 @@ async def process_with_rag(
             # If messages format is provided (for multimodal VLM enhanced query), use it directly
             if messages:
                 return openai_complete_if_cache(
-                    "gpt-4o",
+                    vision_model,
                     "",
                     system_prompt=None,
                     history_messages=[],
@@ -151,7 +154,7 @@ async def process_with_rag(
             # Traditional single image format
             elif image_data:
                 return openai_complete_if_cache(
-                    "gpt-4o",
+                    vision_model,
                     "",
                     system_prompt=None,
                     history_messages=[],


LarFii:
Thanks for your contribution!

I did find one P1 blocker.

1. P1: `example_knowledge_qa.py`'s `--query` flow is broken and will raise at runtime.
   The new script builds a `RAGAnything` instance in `initialize_rag()` ([PR file `example_knowledge_qa.py`, lines 99-116](https://github.com/HKUDS/RAG-Anything/pull/231/files#diff-0)) and then calls `rag.aquery(...)` in the `--query` path ([lines 135-142](https://github.com/HKUDS/RAG-Anything/pull/231/files#diff-0) and [187-192](https://github.com/HKUDS/RAG-Anything/pull/231/files#diff-0)). But upstream `aquery()` explicitly raises when `self.lightrag is None` ([`raganything/query.py`, lines 119-122](https://github.com/HKUDS/RAG-Anything/blob/main/raganything/query.py#L119-L122)). This script never initializes or reloads `LightRAG` before querying, so the advertised workflow (`--process` first, then `--query`) still fails on a fresh process. In practice, users following the example will hit a `ValueError` instead of getting an answer. This should be fixed before merge, either by ensuring the query path initializes/restores `LightRAG` from `WORKING_DIR`, or by using an API that performs that initialization before `aquery()`.

zhangzhenfei:
Thanks for the review! The P1 blocker has been fixed — `initialize_rag()` now calls `await rag._ensure_lightrag_initialized()` before returning, so LightRAG is loaded from `WORKING_DIR` on every code path (including `--query`). The three commits have also been squashed into one.

LarFii:
@zhangzhenfei Hi, taking the liberty to push a small cleanup directly to your branch (you have "Allow edits from maintainers" enabled, thanks!) so we can move this forward:

- rebased onto current `main`
- removed the new top-level `example_knowledge_qa.py` file because it's out of scope for this PR title (env-var model names) and has a few independent issues (wrong location, double `initialize_rag()` in the `--process` branch, `vision_model_func` builds a messages list that can contain `None`, `openai_embed.func` is not a valid attribute on the LightRAG helper). Happy to review it as a separate PR if you'd like to revisit it.
- kept the actual env-var changes in `examples/raganything_example.py` and `env.example`, and preserved your authorship (with `Co-authored-by`) on the squashed commit.

If this looks good I'll merge. Thanks for the contribution!

zhangzhenfei:
Thanks for the cleanup and the detailed explanation! The changes look good to me — feel free to merge. I'll follow up with a separate PR for `example_knowledge_qa.py` after addressing the issues you mentioned. Appreciate the help!

PR #239, opened 2026-04-03
feat: incremental folder scan — skip unchanged files via MD5 manifest

### LINKED ISSUE #156: [Feature Request]: support incremental scan of a folder and update the changed file based on date and md5

- [ ] I have searched the existing feature request and this feature request is not already filed.
- [ ] I believe this is a legitimate feature request, not just a question or bug.

support incremental scan of a folder and update the changed file based on  date and md5

### Additional Context

_No response_

jwchmodx:
## Summary

Closes #156 — adds `incremental=True` to `process_folder_complete` so that unchanged files are skipped on subsequent runs.

### How it works

When `incremental=True` is passed:

1. `process_folder_complete` computes the **MD5 digest** of every discovered file.
2. The digest is compared against a lightweight JSON manifest stored in `config.working_dir` (one manifest per source folder, keyed by a hash of the folder path so multiple folders coexist safely).
3. **Unchanged files** (digest matches the manifest) are silently skipped.
4. **New or changed files** are processed normally.
5. After the run, successfully processed files are marked in the manifest. **Failed files are removed** so they are automatically retried on the next call.

The `incremental=False` default means existing code is completely unaffected.

### New API

```python
await rag.process_folder_complete(
    folder_path="./documents",
    incremental=True,   # skip unchanged files
)
```

### Changes

| File | Change |
|------|--------|
| `raganything/batch.py` | Added `incremental` param + 4 helper methods + manifest update logic |
| `tests/test_incremental_folder_scan.py` | 12 new unit tests |

## Test plan

- [x] `test_processes_all_files_on_first_run` — first run always processes everything
- [x] `test_skips_unchanged_files_on_second_run` — unchanged files are skipped
- [x] `test_reprocesses_changed_file` — modified file is re-processed
- [x] `test_processes_newly_added_file` — new files are picked up
- [x] `test_failed_file_is_retried_next_run` — failed files removed from manifest
- [x] `test_non_incremental_does_not_create_manifest` — no side effects when `incremental=False`
- [x] MD5 helpers + manifest I/O (3 + 3 tests)

All 12 pass; the 4 pre-existing failures in `test_callbacks` and `test_chinese_cid_font` are unrelated to this change.

### DIFF
--- a/raganything/batch.py
+++ b/raganything/batch.py
@@ -5,6 +5,8 @@ Contains methods for processing multiple documents in batch mode
 """
 
 import asyncio
+import hashlib
+import json
 import logging
 from pathlib import Path
 from typing import List, Dict, Any, Optional, TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


LarFii:
Thanks for your contribution!

1. `P1` Stale document versions are kept after a file changes  
   In [`raganything/batch.py:147`](./raganything/batch.py) the new incremental path decides whether to re-run a file, but when a changed file is reprocessed it just calls `process_document_complete()` again. That matters because [`raganything/processor.py:94`](./raganything/processor.py#L94) generates `doc_id` from document content, and [`raganything/processor.py:1556`](./raganything/processor.py#L1556) uses that new content-based ID for insertion. I could not find any cleanup of the previous version before the reinsert. The result is that editing a file will create a new document identity while the old chunks/entities/relations remain in the index, so retrieval can return both old and new versions of the same source file. For an “incremental update” feature, that is a correctness bug, not just a UX gap.

2. `P1` The manifest key ignores processing configuration, so valid reruns are skipped  
   The manifest entry in [`raganything/batch.py:159`](./raganything/batch.py#L159) only checks file-content MD5. If the user changes `parse_method`, parser backend/config, splitting settings, or even points the run at a different output layout, unchanged files are silently skipped and keep results produced under the old configuration. That is inconsistent with the existing parser cache design, which explicitly treats parsing config as part of cache validity in [`raganything/processor.py:164`](./raganything/processor.py#L164). So the new incremental layer can now suppress necessary reprocessing and leave stale outputs/index state behind.

LarFii:
@jwchmodx Thanks for taking on incremental folder scanning, and apologies for the long delay on this one.

Closing this PR after re-reading both P1 items I raised earlier — they are design-level rather than fixable in place, and the codebase has shifted enough since this branch was opened that a clean rework is the right path:

1. **Stale doc versions on file change**: `doc_id` is content-hashed in `processor.py`, so editing a file produces a new doc identity while the previous chunks/entities/relations remain in the index. An incremental layer needs a `path -> last_doc_id` map and an `adelete_by_doc_id` step before reinsert. That's a non-trivial schema addition (manifest format, lifecycle, what to do when the user changes parser config and the previous version becomes unreachable, etc.).

2. **Manifest key ignores parsing config**: skipping based on file MD5 alone silently masks `parse_method` / parser backend / split-config changes. The existing parser cache in `processor.py:164` already treats parsing config as part of cache validity, and the incremental layer needs to mirror that to stay consistent.

Both belong in a fresh design pass that also accounts for the multimodal-status changes that just landed in #255 (separate `multimodal_status` KV namespace, fallback semantics) and the upcoming RAG-Anything ↔ LightRAG integration work in HKUDS/LightRAG#2830, since both shift where document state lives.

If you'd like to revisit this, please reopen as a new PR — I'm happy to discuss a manifest schema and the delete-before-reinsert hook upfront before you write the code. Really appreciate the contribution either way.

PR #240, opened 2026-04-03
fix: prevent crashes from uninitialized LightRAG, env-var stripping, and parser cleanup

jwchmodx:
## Summary

This PR fixes three categories of bugs found in `processor.py`, `config.py`, and `parser.py`.

### Bug fixes

**`processor.py` — LightRAG accessed before initialization (crash risk)**

- `_process_multimodal_content()`: `self.lightrag.doc_status` was accessed at the top of the function (inside a `try/except`) *before* `_ensure_lightrag_initialized()` was called further down. If `self.lightrag` is `None` at that point the `AttributeError` is silently swallowed and the function continues in a broken state. Fix: call `_ensure_lightrag_initialized()` first and return early on failure.
- `process_document_complete_lightrag_api()`: same pattern — `self.lightrag.doc_status.get_by_id(doc_pre_id)` was called on line 1678 before the `try` block that initialises LightRAG on line 1682. Fix: move the `doc_status` fetch after initialisation (it is already fetched again at line 1707, so the early call was redundant).
- `process_document_complete()` and `insert_content_list()`: `_ensure_lightrag_initialized()` return value was ignored (`await self._ensure_lightrag_initialized()`). A failed initialisation would silently fall through and crash later on an unrelated line. Fix: capture and check the result, raising `RuntimeError` on failure.

**`config.py` — missing `.strip()` on env-var splits**

`SUPPORTED_FILE_EXTENSIONS` and `CONTEXT_FILTER_CONTENT_TYPES` are split on `","` without stripping whitespace. A value like `".pdf, .jpg"` produces `[".pdf", " .jpg"]`; the leading space means `" .jpg"` never matches any file extension check. Fix: add a list-comprehension with `.strip()`.

**`parser.py` — redundant `platform.system()` calls and unused `TypeVar`**

`platform.system() == "Windows"` appeared in five separate functions, each preceded by a local `import platform`. This is evaluated at runtime on every call. Fix: hoist it to a single module-level constant `_IS_WINDOWS` and remove the five local imports. Also removed the unused `TypeVar T` import (it was imported and assigned but never referenced).

## Test plan

- [ ] Run existing test suite: `pytest tests/`
- [ ] Verify `RAGAnything` still initializes and processes documents correctly when `lightrag` is pre-provided
- [ ] Verify `RAGAnything` raises a clear error (not a cryptic `AttributeError`) when `llm_model_func` is missing
- [ ] Set `SUPPORTED_FILE_EXTENSIONS=".pdf, .jpg"` (with space) and confirm both extensions are recognized
- [ ] Confirm no regressions on Windows subprocess behavior


LarFii:
Thank you for your contribution! 

I used Codex to review PR #240 against main, focusing on P0 and P1 issues. No P0 issues were found, but one P1 regression was identified.

1. `[P1]` Initialization failures are now silently dropped instead of being persisted as a failed document status. In [`processor.py`](./raganything/processor.py#L1688), `process_document_complete_lightrag_api()` now returns `False` immediately when `_ensure_lightrag_initialized()` fails. The problem is that this return happens before the function ever loads or upserts `doc_pre_id` status, which only starts later at [`processor.py`](./raganything/processor.py#L1707). On `main`, the same failure path explicitly wrote `DocStatus.FAILED` plus `error_msg` before returning. With this PR, parser/bootstrap/storage-init failures become “silent” from the status store’s perspective: callers get `False`, but any polling UI / API that relies on `doc_status` can see a stale or missing record and treat the job as not started rather than failed. That is a real behavior regression for observability and recovery flows, so I would block on restoring the failed-status write before return.

PR #242, opened 2026-04-09
fix: create example log directory correctly

haosenwang1018:
## Summary
- create the actual example log directory instead of its parent in `configure_logging()`
- add a focused regression test covering nested `LOG_DIR` values
- keep the fix scoped to the current mainline example file that replaced the older OpenAI demo path

## Testing
- `uv run python -m pytest tests/test_raganything_example.py -q`
- `uv run ruff check examples/raganything_example.py tests/test_raganything_example.py`
- `uv run ruff format --check examples/raganything_example.py tests/test_raganything_example.py`
- `git diff --check`


LarFii:
Thanks for your contribution!

PR #251, opened 2026-04-21
fix: detect ainsert signature instead of silently swallowing TypeError

### LINKED ISSUE #244: [Bug]: No doc_status created after process_document_complete

### Describe the bug

No kv_store_doc_status is generated after unning process_document_complete. 

I believe there is something related to the **multimodal_content** varible when calling **ainsert**  in your source code. Apparently there is no multimodal_content in the latest branch of lightrag and it leads to a silent failure. 

And our ultimate goal is to run **adelete_by_doc_id** to delete a document. but right now it just returned 404 not found due to no doc_status found

### Steps to reproduce

So i run process_document_complete with these parameters:

```
            file_path=uploaded_file_path,
            output_dir="output",
            parse_method="auto",
            backend="pipeline",
            doc_id=custom_doc_id,
            file_name=data_item.filename,
        )
```

but the doc_status file is not generated as in the image below:

Image

### Expected Behavior

The doc_status should be created 

### LightRAG Config Used

# Paste your config here

### Logs and screenshots

RAGAnything ainsert screenshot:

Image

lightrag ainsert screenshot:

Image

### Additional Information

- LightRAG Version:1.4.14
- Operating System:Windows 10 Pro 22H2 OS build: 19045.6466
- Python Version: Python 3.10.10
- Related Issues:

peterCheng123321:
## Summary

Fixes #244.

`insert_text_content_with_multimodal_content` in `utils.py` was wrapping the entire `lightrag.ainsert()` call in a bare `except Exception` that logged a hint and returned silently. When LightRAG's `ainsert` does not accept a `multimodal_content` keyword argument (i.e. the standard PyPI release rather than the RAGAnything branch), the `TypeError` was swallowed — meaning:

- No text was indexed
- No `kv_store_doc_status` entry was written
- `adelete_by_doc_id` returned 404

**Fix:** use `inspect.signature` to check upfront whether `ainsert` accepts `multimodal_content`. If it does, call with the full argument set. If not, emit a visible `logger.warning` and fall back to text-only insertion so the document is still indexed and `doc_status` is correctly written. Real errors (network failures, LLM errors, etc.) now propagate instead of being silently dropped.

## Test plan
- [ ] Run `process_document_complete()` against the standard PyPI LightRAG release — confirm `kv_store_doc_status` is created and `adelete_by_doc_id` succeeds
- [ ] Run against the RAGAnything LightRAG branch — confirm full multimodal path still works
- [ ] Confirm any real `ainsert` errors (e.g. bad API key) now surface as exceptions

### DIFF
--- a/raganything/utils.py
+++ b/raganything/utils.py
@@ -203,10 +203,14 @@ async def insert_text_content_with_multimodal_content(
         file_paths: single string of the file path or list of file paths, used for citation
         scheme_name: scheme name (optional)
     """
+    import inspect
+
     logger.info("Starting text content insertion into LightRAG...")
 
-    # Use LightRAG's insert method with all parameters
-    try:
+    sig = inspect.signature(lightrag.ainsert)
+    supports_multimodal = "multimodal_content" in sig.parameters
+
+    if supports_multimodal:
         await lightrag.ainsert(
             input=input,
             multimodal_content=multimodal_content,
@@ -216,10 +220,19 @@ async def insert_text_content_with_multimodal_content(
             ids=ids,
             scheme_name=scheme_name,
         )
-    except Exception as e:
-        logger.info(f"Error: {e}")
-        logger.info(
-            "If the error is caused by the ainsert function not having a multimodal content parameter, please update the raganything branch of lightrag"
+    else:
+        logger.warning(
+            "LightRAG.ainsert does not support multimodal_content; "
+            "falling back to text-only insertion. Upgrade to the RAGAnything "
+            "branch of LightRAG to enable full multimodal support."
+        )
+        await lightrag.ainsert(
+            input=input,
+            file_paths=file_paths,
+            split_by_character=split_by_character,
+            split_by_character_only=split_by_character_only,
+            ids=ids,
+            scheme_name=scheme_name,
         )
 
     logger.info("Text content insertion complete")


LarFii:
Thanks for the fix. I checked this against the current LightRAG `ainsert` signature, and I think there is still a blocker before this can be merged.

The fallback path only checks whether `ainsert` supports `multimodal_content`, but it still passes `scheme_name=scheme_name` when `multimodal_content` is not supported. Current upstream/PyPI-style LightRAG `ainsert` also does not accept `scheme_name`, so callers that pass a non-empty scheme will still hit `TypeError: unexpected keyword argument 'scheme_name'` and the document/status creation path can still fail.

Could you please update the compatibility logic to filter all optional extension kwargs against the actual `ainsert` signature, or omit unsupported kwargs entirely in the fallback path? #255 has a more complete version of that direction, so this PR may also be superseded by that one once the remaining doc_status semantics are resolved.

LarFii:
@peterCheng123321 Thanks for the quick turnaround on this. Closing in favor of #255, which has just merged into main. The complete fix for #244 lives there:

- filters all optional `ainsert` kwargs (`multimodal_content` *and* `scheme_name`) against the actual LightRAG signature, so old PyPI builds that don't accept either of them no longer hit `TypeError`
- bootstraps `doc_status` records up front so `adelete_by_doc_id` works even when LightRAG's `ainsert` falls back to text-only
- adds compatibility multimodal completion state in a separate KV namespace for older LightRAG versions whose `DocProcessingStatus` schema rejects `multimodal_processed`
- ships regression tests covering the empty-`ainsert`-signature case this PR was originally targeting

The signature-detection idea you proposed here ended up being the core mechanism in #255, so your direction was right. Really appreciate the contribution.

PR #253, opened 2026-04-21
fix: store multimodal_processed in separate KV namespace to prevent DocProcessingStatus errors

### LINKED ISSUE #91: [Bug]:

### Describe the bug

I’m encountering the following error when trying to process a document:

`DocProcessingStatus.__init__() got an unexpected keyword argument 'multimodal_processed'
`

The document is a Markdown file containing links but no images. 
Has anyone run into this issue before, or found a fix/workaround?

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

### LINKED ISSUE #119: [Bug]:TypeError: DocProcessingStatus.__init__() got an unexpected keyword argument 'multimodal_processed'

### Describe the bug

When I upload PDF then:

500 Internal Server Error {"detail":"DocProcessingStatus.__init__() got an unexpected keyword argument 'multimodal_processed'"} /documents/paginated

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

peterCheng123321:
## Summary

Fixes #91, fixes #119.

RAGAnything was injecting a `multimodal_processed` field directly into LightRAG's `doc_status` KV records. The LightRAG Server API deserializes those records into `DocProcessingStatus` dataclass objects. Any LightRAG version whose `DocProcessingStatus` does not declare `multimodal_processed` raised:

```
DocProcessingStatus.__init__() got an unexpected keyword argument 'multimodal_processed'
500 Internal Server Error on /documents/paginated
```

**Fix:** introduce a dedicated `raganything_multimodal_status` KV namespace (same storage class as `parse_cache`) to hold per-document multimodal processing state. LightRAG's own `doc_status` records are no longer modified with RAGAnything-specific fields, so `DocProcessingStatus` deserialization always succeeds regardless of LightRAG version.

Changes:
- `raganything/raganything.py` — add `multimodal_status` field, initialize in both pre-provided and newly-created LightRAG paths, finalize in `finalize_storages`
- `raganything/processor.py` — all `multimodal_processed` reads/writes now go through `self.multimodal_status`; `doc_status` upserts no longer contain this field

## Test plan
- [ ] Process a document, then call `/documents/paginated` on LightRAG Server — confirm no 500 error
- [ ] Verify `adelete_by_doc_id` still works after processing
- [ ] Confirm `is_document_fully_processed()` and `get_document_processing_status()` return correct values

### DIFF
--- a/raganything/processor.py
+++ b/raganything/processor.py
@@ -531,29 +531,34 @@ class ProcessorMixin:
                 doc_id=doc_id,
             )
 
-        # Check multimodal processing status - handle LightRAG's early DocStatus.PROCESSED marking
+        # Check multimodal processing status using the dedicated multimodal_status store
+        # (kept separate from LightRAG's doc_status so the LightRAG Server API never
+        # encounters unknown fields when deserializing DocProcessingStatus objects)
         try:
+            multimodal_record = (
+                await self.multimodal_status.get_by_id(doc_id)
+                if self.multimodal_status is not None
+                else None
+            )
+            multimodal_processed = (
+                multimodal_record.get("multimodal_processed", False)
+                if multimodal_record
+                else False
+            )
+
+            if multimodal_processed:
+                self.logger.info(
+                    f"Document {doc_id} multimodal content is already processed"
+                )
+                return
+
             existing_doc_status = await self.lightrag.doc_status.get_by_id(doc_id)
             if existing_doc_status:
-                # Check if multimodal content is already processed
-                multimodal_processed = existing_doc_status.get(
-                    "multimodal_processed", False
-                )
-
-                if multimodal_processed:
-                    self.logger.info(
-                        f"Document {doc_id} multimodal content is already processed"
-                    )
-                    return
-
-                # Even if status is DocStatus.PROCESSED (text processing done),
-                # we still need to process multimodal content if not yet done
                 doc_status = existing_doc_status.get("status", "")
                 if doc_status == DocStatus.PROCESSED and not multimodal_processed:
                     self.logger.info(
                         f"Document {doc_id} text processing is complete, but multimodal content still needs processing"
                     )
-                    # Continue with multimodal processing
                 elif doc_status == DocStatus.PROCESSED and multimodal_processed:
                     self.logger.info(
                         f"Document {doc_id} is fully processed (text + multimodal)"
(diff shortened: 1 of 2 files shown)


LarFii:
Thanks for working on this. I reviewed the separate `raganything_multimodal_status` KV namespace approach against the current LightRAG `DocProcessingStatus` model.

I don't think this should be merged as an isolated small fix yet. Moving `multimodal_processed` out of `doc_status` avoids older LightRAG deserialization issues, but it also changes the status semantics for newer LightRAG versions where `DocProcessingStatus.multimodal_processed` is now a first-class field and can drive `PROCESSED` -> `PREPROCESSED` presentation while multimodal processing is incomplete.

Before this can land, I think we need at least:

- a migration or read-fallback plan for existing `doc_status` records that already contain `multimodal_processed`, otherwise `is_document_fully_processed()` can regress until documents are reprocessed;
- tests covering `get_document_processing_status()` / `is_document_fully_processed()` and the LightRAG server pagination failure this is meant to avoid;
- an explicit plan for how this coordinates with #255, because both PRs touch the same `doc_status` / multimodal completion area and will conflict semantically;
- a decision on whether LightRAG server APIs should continue to reflect multimodal completion state, or whether that state becomes RAG-Anything-only metadata.

So I would hold this for a design pass rather than merging it directly.

LarFii:
@peterCheng123321 Thanks for raising this — closing in favor of #255 (now merged into main), which incorporates the core idea you proposed here.

Your direction was right: storing multimodal completion state outside `doc_status` is needed for compatibility with LightRAG versions whose `DocProcessingStatus` dataclass rejects unknown kwargs. #255 ships exactly that, but as a **fallback** rather than a wholesale replacement:

- it still tries to write `multimodal_processed` into `doc_status` first (so newer LightRAG versions that natively support the field keep driving `PROCESSED` -> `PREPROCESSED` presentation as designed in HKUDS/LightRAG#2211 / HKUDS/LightRAG#2248)
- if `doc_status.upsert` rejects the field, it transparently falls back to a separate `multimodal_status` KV namespace (same approach you took here)
- read paths (`is_document_fully_processed`, `get_document_processing_status`, `_process_multimodal_content`'s skip-already-done check) consult both stores via `_get_multimodal_processed_flag`

This avoids the migration concern I raised in the earlier review: existing `doc_status` records that already contain `multimodal_processed` keep working without reprocessing, and #91 / #119's `DocProcessingStatus` deserialization error is also fixed because we no longer write the field unconditionally.

If you spot any case where #255's approach still misbehaves on your setup, please open a new issue with reproduction details — happy to iterate. Really appreciate the contribution.

PR #258, opened 2026-04-22
fix: forward system_prompt parameter in aquery_with_multimodal

### LINKED ISSUE #257: fix: forward system_prompt parameter in aquery_with_multimodal
## Bug Description

The `aquery_with_multimodal` method in `raganything/query.py` lacks an explicit `system_prompt: str | None = None` parameter. While `**kwargs` captures extra arguments, `system_prompt` is never forwarded to the inner `aquery()` calls within the method. This means callers cannot use a custom system prompt with the multimodal variant.

## Steps to Reproduce

```python
result = await rag.aquery_with_multimodal(
    "Describe this image",
    multimodal_content=[{"type": "image", "img_path": "./img.png"}],
    system_prompt="You are an expert image analyst."
)
# system_prompt is silently ignored
```

## Expected Behavior

`system_prompt` should be passed through to all `aquery()` calls inside `aquery_with_multimodal`.

## Fix

Add `system_prompt: str | None = None` to the method signature and explicitly forward it to all inner `aquery(...)` calls.

kuishou68:
## Summary

Fixes a bug where `system_prompt` was silently ignored when calling `aquery_with_multimodal`.

## Changes

- Added `system_prompt: str | None = None` parameter to `aquery_with_multimodal` method signature (before `**kwargs`)
- Forward `system_prompt` to the fallback `aquery()` call when no multimodal content is provided
- Forward `system_prompt` to the main `aquery()` call for the enhanced query
- Updated docstring to document the new parameter

## Why

Without this fix, callers passing `system_prompt` to `aquery_with_multimodal` would have it silently ignored, even though the inner `aquery()` method supports it.

Closes #257

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

### DIFF
--- a/raganything/query.py
+++ b/raganything/query.py
@@ -86,6 +86,7 @@ class QueryMixin:
                 "top_k",
                 "max_tokens",
                 "temperature",
+                "system_prompt",
                 # "only_need_context",
                 # "only_need_prompt",
             ]
@@ -249,7 +250,11 @@ class QueryMixin:
 
         # Generate cache key for multimodal query
         cache_key = self._generate_multimodal_cache_key(
-            query, multimodal_content, mode, **kwargs
+            query,
+            multimodal_content,
+            mode,
+            system_prompt=system_prompt,
+            **kwargs,
         )
 
         # Check cache if available and enabled
--- a/raganything/query.py
+++ b/raganything/query.py
@@ -196,6 +196,7 @@ class QueryMixin:
         query: str,
         multimodal_content: List[Dict[str, Any]] = None,
         mode: str = "mix",
+        system_prompt: str | None = None,
         **kwargs,
     ) -> str:
         """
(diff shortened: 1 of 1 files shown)


Abdeltoto (approved):
Thanks @kuishou68 — clean, surgical fix. Verified the diff:

- Signature change is backward-compatible (`system_prompt` defaults to `None`, placed before `**kwargs`).
- Both forwarding paths are covered (the early text-only fallback and the post-enhanced query call).
- Closes #257 cleanly.

LGTM 👍

One tiny nit (non-blocking): the docstring entry "Optional system prompt to include in the query" could be slightly more precise — something like "Optional system prompt forwarded as-is to the underlying `aquery` (see LightRAG's `aquery` for exact handling)" — to make it clear there's no transformation happening here. Totally optional, feel free to ignore.

kuishou68:
Good call on the docstring — updated it to clarify that `system_prompt` is forwarded as-is to the underlying `aquery` without transformation. Makes the contract clearer.

PR #261, opened 2026-04-22
refactor(parser): replace Docling CLI subprocess with Python API (closes #222)

### LINKED ISSUE #222: [Feature Request]: Replace Docling Parser's CLI subprocess with Python API

## Summary 

- Replace the current `DoclingParser` implementation that shells out to the
`docling` CLI via `subprocess.run` with a direct integration through the
[Docling Python API](https://github.com/DS4SD/docling)
(`docling.document_converter.DocumentConverter`).
 
This eliminates process-spawning overhead, avoids disk I/O round-trips for
intermediate JSON/Markdown files, and enables in-memory model reuse across
consecutive parse calls — yielding significant performance gains for
multi-document workloads while preserving full backward compatibility.
 
---
 
## Motivation / Problem
 
The current `DoclingParser` in HKUDS/RAG-Anything invokes Docling through
its command-line interface:
 
```python
# Current approach (raganything/parser.py – _run_docling_command)
cmd = [
"docling",
"--output", str(file_output_dir),
"--to", "json",
"--to", "md",
str(input_path),
]
result = subprocess.run(cmd, **docling_subprocess_kwargs)
```
 
- After the subprocess completes, output files are read back from disk
(`_read_output_files`). This pattern has several drawbacks:
 
\| Issue \| Impact \|
\|-------\|--------\|
\| **Process-spawn overhead** \| Each `parse_*` call forks a new process, loads the Python interpreter, and re-initializes all Docling models from scratch. \|
\| **Disk I/O round-trip** \| Docling writes JSON + Markdown to disk; the parser then reads them back. This is unnecessary when the data is immediately consumed in-memory. \|
\| **No model reuse** \| Docling's deep-learning models (table structure, OCR, layout) are loaded fresh on every invocation — the most expensive part of the pipeline. \|
\| **Fragile error handling** \| Errors surface as `subprocess.CalledProcessError` with stderr strings rather than typed Python exceptions with full stack traces. \|
\| **Platform quirks** \| Windows requires `CREATE_NO_WINDOW` flags; PATH must include the `docling` entry-point. These are unnecessary when calling Python directly. \|
\| **Limited pipeline control** \| CLI flags expose only a subset of Docling's configuration surface. The Python API offers fine-grained control over pipeline options, format options, and OCR settings. \|
 
---
 
### Core Idea
 
Use `docling.document_converter.DocumentConverter` directly in Python:
 
```python
from docling.document_converter import DocumentConverter, PdfFormatOption
from docling.datamodel.base_models import InputFormat
from docling.datamodel.pipeline_options import PdfPipelineOptions, TableFormerMode
 
# Build pipeline options from user kwargs
pipeline_options = PdfPipelineOptions()
pipeline_options.table_structure_options.mode = TableFormerMode.ACCURATE
pipeline_options.do_ocr = True
 
# Create converter (reused across calls)
converter = DocumentConverter(
format_options={
InputFormat.PDF: PdfFormatOption(pipeline_options=pipeline_options),
}
)
 
# Parse document — no subprocess, no disk I/O
result = converter.convert(str(input_path))
doc_dict = result.document.export_to_dict()
 
# Convert to MinerU-compatible content-list format
content_list = self.read_from_block_recursive(
doc_dict["body"], "body", img_output_dir, 0, "0", doc_dict
)
```
 ---
 
## Backward Compatibility
 
- **No breaking changes** to the public `parse_pdf()`, `parse_document()`,
`parse_office_doc()`, `parse_html()` signatures.
- Callers passing `env={"KEY": "VAL"}` (previously used for subprocess
environment) will have the type validated but the value silently ignored.
 
### Output Format
 
- The content-list output format is **identical** to the current
implementation. The same `read_from_block_recursive()` /
`read_from_block()` methods are used to transform the Docling document
dict into the MinerU-compatible structure.
 
---
 
### Dependencies
 
- **No new required dependencies.** `docling` remains an optional package.
- The `check_installation()` method gracefully reports whether the package
is available.
- `pip install docling` is the only setup step for users who want to use
the Docling parser backend.
 
---
 
### For End Users
 
No action required. The public API is unchanged:
 
```python
from raganything.parser import DoclingParser
 
parser = DoclingParser()
content = parser.parse_document("report.pdf", output_dir="./output")
```
 
### For Callers Passing `env`
 
The `env` kwarg is still accepted but no longer has any effect:
 
```python
# Before (CLI subprocess):
parser.parse_pdf("doc.pdf", env={"DOCLING_CACHE": "/tmp"})
# After (Python API): accepted without error, but env is ignored.
```
 
### For Advanced Configuration
 
The Python API exposes more options than the CLI:
 
```python
parser.parse_pdf(
"doc.pdf",
table_mode="accurate",      # TableFormerMode.ACCURATE
tables=True,                # do_table_structure = True
allow_ocr=True,             # do_ocr = True
artifacts_path="/models",   # custom model artifacts directory
)
```
 
---
 
## Checklist
 
- [x] Backward compatibility preserved for all existing kwargs
- [x] Output format identical to current CLI-based implementation
- [x] No new required dependencies introduced 
---
 
## References
 
- Docling Python API: https://github.com/DS4SD/docling
- `DocumentConverter` usage: https://ds4sd.github.io/docling/

### Additional Context

_No response_

Abdeltoto:
## Summary

Closes #222.

Replaces `DoclingParser`'s subprocess-based path that shells out to the `docling` CLI on every parse call with a direct integration through the Docling Python API (`docling.document_converter.DocumentConverter`). Eliminates process-spawn overhead, removes the JSON disk round-trip, and — most importantly — enables in-memory model reuse across consecutive parse calls via a per-pipeline-option converter cache.

## What changed

- **New `_get_converter(**kwargs)`**: lazily builds a `DocumentConverter` and caches one instance per `(table_mode, do_tables, do_ocr, artifacts_path)` tuple, so layout / OCR / TableFormer model weights load **once per process** for a given configuration.
- **New `_run_docling_python(...)`**: invokes the cached converter, exports the document via `result.document.export_to_dict()`, still writes the legacy `<file_stem>.json` and `<file_stem>.md` artifacts to `<output_dir>/<file_stem>/docling/` for backward compatibility, and returns the in-memory dict.
- **`parse_pdf` / `parse_office_doc` / `parse_html`** now feed the in-memory dict directly to `read_from_block_recursive` — the JSON disk round-trip is gone.
- **`check_installation()`** switches from `subprocess.run(["docling", "--version"])` to `import docling.document_converter`, which is faster, more accurate (it tests the actual import path the parser uses), and removes the Windows `CREATE_NO_WINDOW` quirk.
- **The legacy `env={...}` kwarg** is still accepted and type-validated for backward compatibility, but now logs a debug message and is otherwise ignored — the Python API path does not require subprocess environment overrides.

## Backward compatibility

- Public signatures of `parse_pdf`, `parse_office_doc`, `parse_html`, `parse_document`, and `check_installation` are unchanged.
- The on-disk layout (`<output_dir>/<file_stem>/docling/<file_stem>.json` and `.md`) is preserved — anything that grepped those files keeps working.
- Image extraction continues to write PNGs into `<file_stem>/docling/images/` via the existing `read_from_block` logic.
- Picture image data is now requested from the converter via `generate_picture_images=True` so that base64 picture URIs are present in the dict, mirroring what the CLI produced.
- No new required dependencies. `docling` remains an optional install (`pip install docling`).

## Performance

The big wins are not visible on a single document — they show up on multi-document workloads where the cached converter is reused:

1. **No subprocess fork** per call.
2. **No Python interpreter re-init** per call.
3. **Layout / OCR / TableFormer models loaded once** instead of N times.
4. **No JSON disk round-trip** between parsing and content-list construction.

For workloads that run dozens of `.pdf` / `.docx` ingestions back-to-back this is a substantial speedup — exactly the case that was painful with the CLI-based path.

## Test plan

- [x] `ruff format` and `ruff check --ignore=E402` pass on `raganything/parser.py`.
- [x] AST parses cleanly and no lints reported.
- [ ] Maintainer-side: end-to-end run on a representative `.pdf` and `.docx` to confirm the produced `content_list` is byte-identical (or at least equivalent) to the previous CLI-based output, and that the `<file_stem>.json` / `.md` artifacts on disk look right.

Happy to iterate on naming, the `_get_converter` cache key shape, or whether we should keep writing the legacy on-disk JSON / Markdown at all (current PR keeps it for safety).

### DIFF
--- a/raganything/parser.py
+++ b/raganything/parser.py
@@ -32,6 +32,7 @@ import argparse
 import base64
 import subprocess
 import tempfile
+import threading
 import logging
 import time
 import urllib.parse
@@ -1459,6 +1460,33 @@ class DoclingParser(Parser):
     models on every call. A `DocumentConverter` instance is built lazily on first
     use and cached per pipeline-option combination so that subsequent parses
     against the same configuration reuse already-loaded models.
+
+    Compatibility changes vs. earlier CLI-subprocess implementation
+    ----------------------------------------------------------------
+    - `check_installation()` now returns True iff the Docling Python package
+      can be imported (`docling.document_converter.DocumentConverter`). The
+      previous behavior of probing the `docling` CLI executable on PATH is
+      gone; environments that ship the CLI without the importable package
+      (or vice versa) will see a different result than before.
+    - The legacy `env={...}` kwarg is still accepted for source-level
+      compatibility but is **ignored**: the Python API does not run a
+      subprocess, so per-call environment overrides no longer take effect.
+      Callers needing model-cache, proxy, or CUDA configuration should set
+      the corresponding environment variables in the parent process before
+      instantiating `DoclingParser`, or configure Docling directly via
+      `_get_converter` kwargs (`artifacts_path`, `table_mode`, ...).
+    - JSON and Markdown artifacts are still written to
+      `<output_dir>/<file_stem>/docling/` for backward compatibility, but
+      they are produced by Docling's `export_to_dict()` /
+      `export_to_markdown()` rather than by the CLI's serializer; expect the
+      same logical content but not byte-identical files (key ordering,
+      whitespace, optional fields may differ).
+
+    Concurrency
+    -----------
+    The internal converter cache is guarded by a lock so that a single
+    `DoclingParser` instance can be safely shared across threads without
+    duplicating Docling model loads on first use.
     """
 
     # Define Docling-specific formats
@@ -1469,7 +1497,11 @@ class DoclingParser(Parser):
         super().__init__()
         # Cache of DocumentConverter instances keyed by pipeline-option tuple,
         # so that loaded layout/OCR/table models are reused across calls.
+        # The lock guards concurrent first-use from creating duplicate
+        # converters (and re-loading models) when the same DoclingParser
+        # instance is shared across threads.
         self._converter_cache: Dict[Tuple, Any] = {}
+        self._converter_cache_lock = threading.Lock()
 
     def parse_pdf(
         self,
@@ -1616,6 +1648,9 @@ class DoclingParser(Parser):
         artifacts_path = kwargs.get("artifacts_path")
 
         cache_key = (table_mode, do_tables, do_ocr, artifacts_path)
+        # Fast path: snapshot read outside the lock (dict reads are atomic in
+        # CPython for hashable keys) so the common cache-hit case stays
+        # contention-free.
         cached = self._converter_cache.get(cache_key)
         if cached is not None:
             return cached
@@ -1645,13 +1680,23 @@ class DoclingParser(Parser):
         if hasattr(pipeline_options, "images_scale"):
             pipeline_options.images_scale = 2.0
 
-        converter = DocumentConverter(
-            format_options={
-                InputFormat.PDF: PdfFormatOption(pipeline_options=pipeline_options),
-            }
-        )
-        self._converter_cache[cache_key] = converter
-        return converter
+        # Slow path: serialize converter construction so that concurrent
+        # first-use against the same cache_key doesn't load Docling's models
+        # twice. We re-check the cache under the lock to avoid a double build
+        # when two threads race past the fast-path check above.
+        with self._converter_cache_lock:
+            cached = self._converter_cache.get(cache_key)
+            if cached is not None:
+                return cached
+            converter = DocumentConverter(
+                format_options={
+                    InputFormat.PDF: PdfFormatOption(
+                        pipeline_options=pipeline_options
+                    ),
+                }
+            )
+            self._converter_cache[cache_key] = converter
+            return converter
 
     def _run_docling_python(
         self,
@@ -1980,6 +2025,15 @@ class DoclingParser(Parser):
         Returns:
             bool: True if `docling.document_converter.DocumentConverter` can be
                 imported, False otherwise.
+
+        Note:
+            This is a behavior change from the previous CLI-subprocess
+            implementation, which probed the `docling` executable on PATH.
+            Some environments may have the CLI installed without the Python
+            package (or vice versa) and will therefore see a different
+            result. The Python-API path is what `parse_pdf`,
+            `parse_office_doc`, and `parse_html` actually exercise, so this
+            check is now a faithful pre-flight for those entry points.
         """
         try:
             from docling.document_converter import DocumentConverter  # noqa: F401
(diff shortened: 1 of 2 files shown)


LarFii:
Thanks for working on this. Replacing the Docling CLI subprocess with the Python API and caching `DocumentConverter` is a good direction, but I think this needs changes before merge.

1. P0: the existing parser kwargs tests still target the subprocess path (`_run_docling_command` / `subprocess.run`). This PR removes that path, so those tests need to be updated to mock `DocumentConverter`, `convert()`, and `export_to_dict()` / `export_to_markdown()` instead. Otherwise CI will regress.

2. P1: `check_installation()` changes from testing the `docling` CLI executable to testing Python importability. That is a real behavior change and should be documented. Some environments can import the package without the CLI, or vice versa.

3. P1: the legacy `env={...}` kwarg is now validated but ignored. Previously it was passed to the subprocess and could be used for proxy/model/cache/CUDA environment overrides. Please call this out as a compatibility change, or provide an equivalent path if Docling still needs those runtime env values.

4. P1: please attach at least one small PDF and one DOCX/HTML validation comparing the produced `content_list` and optional on-disk json/md artifacts against the current CLI behavior. Byte-identical artifacts may not be required, but the expected compatibility level should be stated.

5. P2: if one `DoclingParser` instance can be used concurrently, the converter cache should either be protected or documented as best-effort, since concurrent first-use can create duplicate converters.

Holding this for tests and compatibility notes; happy to re-review after those are added.

Abdeltoto:
Thanks @LarFii — addressed all five (and rebased onto current `main`):

1. **(P0)** `testparser_kwargs.py` now exercises the Python-API path. The Docling tests mock `DoclingParser._get_converter` to return a `DocumentConverter` stub whose `convert()` returns a document with `export_to_dict()` and `export_to_markdown()`. Added coverage for the cache (same-kwargs reuse, distinct-kwargs build).
2. **(P1)** `check_installation()` docstring now explicitly calls out the behavior change: it tests Python importability, not CLI presence on PATH.
3. **(P1)** Class docstring documents that `env={...}` is accepted-but-ignored, with the suggested workaround (set vars in the parent process, or use `_get_converter` kwargs like `artifacts_path` / `table_mode`).
4. **(P1)** Same docstring states the artifact-compatibility level: same logical `content_list`; on-disk JSON/MD come from `export_to_dict()` / `export_to_markdown()` rather than the CLI's serializer, so not byte-identical (key ordering, whitespace, optional fields can differ). I don't have a side-by-side validation environment with both the legacy CLI and a matching Docling version installed; happy to address concrete discrepancies if you spot any during your re-review.
5. **(P2)** Added a `threading.Lock` around `_converter_cache` in `_get_converter`. Fast path stays lock-free; the lock only serializes the build path with a double-checked cache lookup so concurrent first-use can't load the models twice.

8 unit tests pass locally.

PR #262, opened 2026-04-22
feat(omml): add OMML equation extraction utility for DOCX documents (closes #259)

### LINKED ISSUE #259: [Feature Request]: Extract OMML equations from DOCX before parser converts them away

## Summary

Word stores inline and display math as `<m:oMath>` (OMML — Office Math Markup Language) elements inside `word/document.xml`. When RAG-Anything ingests a `.docx` file, both the MinerU and Docling parsers convert the document to PDF first (LibreOffice for MinerU, native conversion for Docling). During that conversion, OMML equations are typically rasterized to images or replaced with placeholder glyphs — the **structured math content is lost** before it ever reaches the multimodal pipeline.

This silently degrades retrieval quality for technical documents (academic papers, lecture notes, engineering reports) — exactly the workloads where structured math matters most. A user who ingests a `.docx` thesis containing 50 equations gets a knowledge graph with **zero** equation entities, even though `RAGAnything` already has a fully working `EquationModalProcessor` waiting for them.

---

## Reproduction

1. Take any `.docx` containing inline or display math (a Word equation editor formula, e.g. `\frac{a}{b}` or `\sum_{i=1}^{n} x_i`).
2. Run `await rag.process_document_complete("paper.docx")` with either parser.
3. Inspect the resulting `content_list`: equation blocks (`{"type": "equation", ...}`) are absent or replaced with image placeholders that contain only the rendered glyph, not the LaTeX source.
4. Compare to the same document fed in directly as `.tex` or as a PDF that was authored in LaTeX — those retain searchable equation content.

---

## Proposed Solution

Add a small, **zero-dependency, pure-stdlib** utility module `raganything/omml_extractor.py` that:

1. Opens the DOCX as a ZIP archive (using `zipfile`).
2. Parses `word/document.xml` (using `xml.etree.ElementTree`).
3. Walks every `<m:oMath>` element in document order.
4. Converts each to LaTeX via a recursive transformer covering the most common OMML constructs (runs, fractions, super/subscripts, radicals, n-ary operators with the standard sum/prod/integral characters, applied functions, delimiters, matrices, accents, limits, grouping characters, phantoms).
5. Returns a list of equation dicts in the **exact MinerU-compatible shape** already consumed by `EquationModalProcessor`:
   ```python
   {"type": "equation", "text": "<latex>", "text_format": "latex", "page_idx": ...}
   ```

A second helper `enrich_content_list_with_docx_equations(content_list, docx_path)` would let users / downstream code merge the extracted equations back into a parsed content list (with deduplication against any equation blocks the parser already produced).

### Why zero new dependencies?

- `zipfile` and `xml.etree` are stdlib.
- No `python-docx` (incomplete OMML support).
- No `lxml` / XSLT (heavyweight, build issues on some platforms).
- No `pandoc` (external binary).

This keeps the change install-friendly and easy to audit.

### Coverage trade-off

The recursive converter handles the **high-frequency cases** observed in academic and engineering documents. Rare constructs (e.g. `m:eqArr` equation arrays, exotic accents) gracefully fall back to text concatenation rather than raising. **For RAG indexing, recall beats fidelity** — even a partially correct LaTeX rendering makes the math searchable, which is a strict improvement over the current "math is invisible" baseline.

---

## Backward Compatibility

- New module — no changes to existing parsers or processors.
- Opt-in: callers explicitly invoke `extract_omml_equations()` or `enrich_content_list_with_docx_equations()`.
- Output dicts use the same schema as existing equation blocks, so they slot into `EquationModalProcessor` without code changes anywhere else.

---

## Tests

A companion test file (`tests/test_omml_extractor.py`) would cover:

- Direct OMML → LaTeX conversion of every supported construct.
- End-to-end extraction from in-memory DOCX archives built with `zipfile`.
- Enrichment behavior (append, deduplicate, copy semantics, empty input).
- Error handling for missing files, invalid ZIPs, and DOCX archives that lack `word/document.xml`.

All tests would pass with the stdlib alone.

---

## Out of Scope (possible follow-ups)

- Positional placement: DOCX paragraphs do not carry page numbers, so the first iteration appends extracted equations at the tail of the content list with the last block's `page_idx`. Precise positional weaving could be a follow-up using `raw_omml` as a join key against parser-produced image placeholders.
- Auto-integration into `process_document_complete()` for `.docx` inputs (deliberately punted to a second PR so the utility lands first and can be reviewed in isolation).

---

## Checklist

- [x] No new required dependencies introduced
- [x] Backward compatible (new module, no changes to existing APIs)
- [x] Output schema identical to existing equation blocks

I have a working implementation with 26 passing tests ready to submit as a PR — opening this issue first so it can be discussed and tracked.

### Additional Context

_No response_

Abdeltoto:
## Summary

Closes #259.

Adds `raganything/omml_extractor.py`, a zero-dependency, pure-stdlib utility that extracts OMML (Office Math Markup Language) equations from `.docx` files and converts them to LaTeX, so the structured math content survives the DOCX → PDF conversion that both the MinerU and Docling parsers rely on.

## Why

Word stores inline and display math as `<m:oMath>` elements inside `word/document.xml`. When a DOCX is ingested today, both parsers convert it to PDF first (LibreOffice for MinerU, native conversion for Docling). During that conversion, equations are typically rasterized to images or replaced with placeholder glyphs — the structured math text is **lost** before it ever reaches `EquationModalProcessor`. This silently degrades retrieval quality for academic papers, lecture notes, and engineering reports — exactly the workloads where structured math matters most.

See #259 for the full motivation and the proposed approach.

## What this PR adds

Two public functions and one transformer:

- **`extract_omml_equations(docx_path)`** — opens the DOCX as a ZIP, parses `word/document.xml` with `xml.etree`, walks every `<m:oMath>` element in document order, and returns a list of dicts with `text` (LaTeX), `text_format`, `index`, and `raw_omml` (the source XML, useful for callers that want to plug in their own converter).

- **`enrich_content_list_with_docx_equations(content_list, docx_path)`** — takes a parsed MinerU-compatible content list and appends `{"type": "equation", "text": "<latex>", "text_format": "latex", ...}` blocks for each extracted equation. Deduplicates against equations already present in the list (e.g. those the parser already produced as image placeholders) by default.

- **`omml_to_latex(element)`** — recursive transformer covering the most common OMML constructs: runs (`m:r`/`m:t`), fractions (`m:f`), super/subscripts (`m:sSup`/`m:sSub`/`m:sSubSup`/`m:sPre`), radicals (`m:rad`), n-ary operators (`m:nary` with the standard sum/prod/integral characters), applied functions (`m:func`, with whitelist for `\sin`, `\cos`, `\log`, etc.), delimiters (`m:d`), matrices (`m:m`), accents (`m:acc`, `m:bar`), limits (`m:limLow`/`m:limUpp`), grouping characters (`m:groupChr`), and phantoms (`m:phant`). Unknown elements gracefully fall back to their text content rather than raising — for RAG indexing, **recall beats fidelity**.

## Tests

`tests/test_omml_extractor.py` covers **26 cases**, all passing locally:

- Direct OMML → LaTeX conversion for every supported construct (fractions, sup/sub, radicals, summation with bounds, integrals with default operator, known and unknown functions, delimiters with default and explicit braces, 2×2 matrices, overline/underline, Unicode symbol substitution).
- End-to-end extraction from in-memory DOCX archives built with `zipfile` (single equation, multiple equations preserve document order).
- Enrichment behavior: append, deduplicate against existing equations, opt-out of dedup, empty input returns a defensive copy.
- Error handling: missing file → `FileNotFoundError`, non-ZIP → `ValueError`, ZIP missing `word/document.xml` → `ValueError`.

All 26 tests pass with the stdlib alone — no `python-docx`, no `lxml`, no `pandoc`.

## Backward compatibility

- New module — **no changes to existing parsers or processors**.
- Strictly opt-in: callers explicitly invoke `extract_omml_equations()` or `enrich_content_list_with_docx_equations()`.
- Output dicts use the same schema as existing equation blocks, so they slot into `EquationModalProcessor` without code changes anywhere else.
- No new required dependencies.

## Out of scope (possible follow-ups)

- **Positional placement**: DOCX paragraphs do not carry page numbers, so this PR appends extracted equations at the tail of the content list with the last block's `page_idx`. Precise positional weaving could be a follow-up using `raw_omml` as a join key against parser-produced image placeholders.
- **Auto-integration into `process_document_complete()`** for `.docx` inputs — deliberately punted to a second PR so the utility lands first and can be reviewed in isolation.

## Test plan

- [x] All 26 unit tests pass locally on Python 3.12 / Windows.
- [x] `ruff format` and `ruff check --ignore=E402` pass on the new files.
- [ ] Maintainer-side: try it on a real `.docx` with mixed inline and display math and confirm the LaTeX is searchable inside the resulting knowledge graph.

Note: the test file is added with `git add -f` because `.gitignore` includes a broad `test_*` pattern that would otherwise hide it; the existing `tests/test_*.py` files are tracked the same way, so this matches the established convention.

### DIFF
--- a/raganything/omml_extractor.py
+++ b/raganything/omml_extractor.py
@@ -1,0 +1,729 @@
+"""
+OMML (Office Math Markup Language) equation extraction for DOCX documents.
+
+Word stores equations as ``<m:oMath>`` elements inside ``word/document.xml``.
+When a DOCX is converted to PDF (e.g. via LibreOffice for the MinerU parser
+or via Docling's native pipeline), inline math is typically rasterized to an
+image or replaced with a placeholder, so the structured math content is lost
+to the downstream RAG pipeline.
+
+This module provides a zero-dependency, pure-stdlib utility to:
+
+1. Open a DOCX as a ZIP archive,
+2. Locate every ``<m:oMath>`` / ``<m:oMathPara>`` element in document order,
+3. Convert each one to a LaTeX string using a recursive transformer that
+   handles the most common OMML constructs (runs, fractions, scripts,
+   radicals, n-ary operators, functions, delimiters, matrices, accents,
+   and grouping characters), and
+4. Optionally merge the extracted equations into an already-parsed
+   ``content_list`` (the MinerU-compatible list of content blocks produced
+   by :class:`raganything.parser.MineruParser` / :class:`DoclingParser`)
+   as ``{"type": "equation", "text": "<latex>", "text_format": "latex", ...}``
+   blocks so that downstream multimodal processors can index them.
+
+The LaTeX produced by :func:`omml_to_latex` is intended to be searchable and
+human-readable rather than typographically perfect; for unhandled constructs
+the converter falls back to the concatenated text content of the element
+rather than failing. This is deliberate: the priority for RAG is recall, not
+pixel-accurate rendering.
+
+Example
+-------
+>>> from raganything.omml_extractor import extract_omml_equations
+>>> equations = extract_omml_equations("paper.docx")  # doctest: +SKIP
+>>> for eq in equations:                              # doctest: +SKIP
+...     print(eq["index"], eq["text"])
+0 \\frac{a}{b}
+1 \\sum_{i=1}^{n} x_{i}^{2}
+"""
+
+from __future__ import annotations
+
+import logging
+import zipfile
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Union
+from xml.etree import ElementTree as ET
+
+logger = logging.getLogger(__name__)
+
+# --- XML namespaces used in WordprocessingML / OMML ----------------------
+# These are the namespace URIs declared by every Word document part. We use
+# them to construct fully qualified tag names so that lookups are robust to
+# arbitrary namespace prefix choices in the source XML.
+NS = {
+    "m": "http://schemas.openxmlformats.org/officeDocument/2006/math",
+    "w": "http://schemas.openxmlformats.org/wordprocessingml/2006/main",
+}
+
+_M = "{" + NS["m"] + "}"
+_W = "{" + NS["w"] + "}"
+
+# OMML "n-ary" operator characters and their LaTeX equivalents. Word stores
+# the operator as a Unicode codepoint inside m:naryPr/m:chr; if the chr
+# attribute is absent it defaults to the integral sign (\u222b).
+_NARY_OPERATORS: Dict[str, str] = {
+    "\u2211": r"\sum",  # SUMMATION
+    "\u220f": r"\prod",  # PRODUCT
+    "\u2210": r"\coprod",  # COPRODUCT
+    "\u222b": r"\int",  # INTEGRAL
+    "\u222c": r"\iint",  # DOUBLE INTEGRAL
+    "\u222d": r"\iiint",  # TRIPLE INTEGRAL
+    "\u222e": r"\oint",  # CONTOUR INTEGRAL
+    "\u222f": r"\oiint",  # SURFACE INTEGRAL
+    "\u22c3": r"\bigcup",  # N-ARY UNION
+    "\u22c2": r"\bigcap",  # N-ARY INTERSECTION
+    "\u2a00": r"\bigodot",
+    "\u2a01": r"\bigoplus",
+    "\u2a02": r"\bigotimes",
+}
+
+# A small mapping of Unicode math symbols Word likes to use in m:t runs.
+# The set is intentionally minimal; anything not in the table is passed
+# through unchanged so that ``\alpha`` and ``α`` both remain searchable.
+_SYMBOL_TO_LATEX: Dict[str, str] = {
+    "\u00b1": r"\pm",
+    "\u00d7": r"\times",
+    "\u00f7": r"\div",
+    "\u2202": r"\partial",
+    "\u2207": r"\nabla",
+    "\u221a": r"\sqrt{}",  # bare radical sign (rare; m:rad handles real ones)
+    "\u221e": r"\infty",
+    "\u2248": r"\approx",
+    "\u2260": r"\neq",
+    "\u2264": r"\leq",
+    "\u2265": r"\geq",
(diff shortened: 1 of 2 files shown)


LarFii:
Thanks for the detailed implementation. The opt-in utility direction makes sense, but I found a few issues that should be addressed or explicitly documented before merge.

1. P1: several OMML conversion paths can crash on malformed or partial OMML. For example, fraction/script/radical helpers call child conversion even when a required child like `num` or `e` is missing. Since the extractor is intended to prefer recall and tolerate imperfect documents, these should return an empty string / fallback text rather than raising from `_convert_children(None)`.

2. P1: unknown `m:nary` operators appear to fall back to `\int`. That silently turns an unsupported operator into an integral, which is worse than preserving the original Unicode/operator text for retrieval. Please preserve the unknown operator or use an explicit unknown placeholder instead of defaulting to integral.

3. P1/P2: `deduplicate_existing_equations` is exact-text-only. That is fine for an opt-in utility, but it will not deduplicate parser-produced formula placeholders or image-derived equations from Docling/MinerU. Please document that limitation clearly, or add a test showing the expected behavior when existing equation blocks contain non-LaTeX placeholder text.

4. P2: all appended equations inherit the last content block's `page_idx` and are appended at the end. Please document this as a known limitation so users do not treat page-level filtering/order as precise for the enriched equations.

The overall approach is useful, and I would be happy to re-review after the crash path and unknown n-ary fallback are fixed or covered by tests/limitations.

Abdeltoto:
Thanks @LarFii — addressed all four:

1. `_convert_children` now tolerates `None`, so fraction/script/radical handlers degrade to `""` on missing children instead of crashing.
2. `_h_nary` preserves the original Unicode character when `m:chr` is unknown (was silently rewritten to `\int`). Default-to-`\int` kept for the no-`chr` case per ECMA-376 §22.1.2.74.
3. & 4. Added a Limitations block to the enrich helper docstring covering exact-text dedup scope and the tail-append / inherited `page_idx`.

A few small tests added for the new fallbacks.

PR #264, opened 2026-04-22
feat: add MiniMax provider support

octo-patch:
## Summary

- Add `examples/minimax_integration_example.py` showing how to use MiniMax as the LLM backend for RAG-Anything, following the same pattern as the existing Ollama and LM Studio examples
- Support **MiniMax-M2.7** and **MiniMax-M2.7-highspeed** models via MiniMax's OpenAI-compatible API (`https://api.minimax.io/v1`)
- Enforce MiniMax's temperature constraint `(0.0, 1.0]` (defaults to `1.0`)
- Add unit tests (`tests/test_minimax_integration.py`) covering configuration, temperature enforcement, model/URL defaults, and API parameter passing

## Usage

```python
from raganything import RAGAnything, RAGAnythingConfig
from lightrag.llm.openai import openai_complete_if_cache, openai_embed
from lightrag.utils import EmbeddingFunc

async def minimax_llm_model_func(prompt, system_prompt=None, history_messages=[], **kwargs):
    kwargs.setdefault("temperature", 1.0)  # MiniMax requires (0.0, 1.0]
    return await openai_complete_if_cache(
        model="MiniMax-M2.7",
        prompt=prompt,
        system_prompt=system_prompt,
        history_messages=history_messages,
        base_url="https://api.minimax.io/v1",
        api_key=os.getenv("MINIMAX_API_KEY"),
        **kwargs,
    )
```

Or run the complete example:

```bash
export MINIMAX_API_KEY=your-api-key
python examples/minimax_integration_example.py
```

## Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `MINIMAX_API_KEY` | — | MiniMax API key (required) |
| `MINIMAX_BASE_URL` | `https://api.minimax.io/v1` | API base URL |
| `MINIMAX_LLM_MODEL` | `MiniMax-M2.7` | Model ID |

> **Note**: MiniMax does not provide an embedding model. A separate embedding service (e.g. OpenAI, Ollama) is required for the embedding function.

## API Reference

- Chat (OpenAI Compatible): https://platform.minimax.io/docs/api-reference/text-openai-api

### DIFF
--- a/examples/minimax_integration_example.py
+++ b/examples/minimax_integration_example.py
@@ -36,17 +36,18 @@ API Reference:
 import os
 import uuid
 import asyncio
-from typing import List, Dict, Optional
+import inspect
+from typing import Dict, List, Optional
 
 from dotenv import load_dotenv
-
-# Load environment variables
-load_dotenv()
 
 # RAG-Anything imports
 from raganything import RAGAnything, RAGAnythingConfig
 from lightrag.utils import EmbeddingFunc
 from lightrag.llm.openai import openai_complete_if_cache, openai_embed
+
+# Load environment variables
+load_dotenv()
 
 # MiniMax configuration
 MINIMAX_BASE_URL = os.getenv("MINIMAX_BASE_URL", "https://api.minimax.io/v1")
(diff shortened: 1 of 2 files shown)


LarFii:
Thanks for the contribution. I reviewed the example and tests, and I think this needs changes before it can be merged.

1. P0: `examples/minimax_integration_example.py` is currently not valid Python. The raw file shows function/class bodies indented with only one leading space, and nested blocks like the temperature guard have no deeper indentation. This will raise `IndentationError` / `SyntaxError`, so the example and the import-based tests cannot run as-is.

2. P1: `test_connection()` relies on `AsyncOpenAI.models.list()`. Many OpenAI-compatible providers either do not implement `/v1/models` or expose it inconsistently, so a valid MiniMax chat-completions setup may be rejected before the actual chat path is tested. Please make this best-effort or use the chat completion test as the primary connectivity check.

3. P1: when `MINIMAX_API_KEY` is missing, the lower LightRAG OpenAI helper can fall back to `OPENAI_API_KEY` if the example path is called directly. Please guard the actual LLM function too, or make the error mode explicit so users do not accidentally send an OpenAI key to the MiniMax base URL.

4. P2: please clarify that this PR adds an OpenAI-compatible example, not a core `LLM_BINDING` provider, and consider adding `MINIMAX_*` variables to `env.example` or a README pointer for discoverability.

Holding off on merge until the syntax issue is fixed and the connectivity behavior is clarified.

PR #287, opened 2026-05-22
docs: add API reference

Kartalops:
## Summary
Adds API reference documentation.

## Testing
N/A.

### DIFF
--- a/docs/api_reference.md
+++ b/docs/api_reference.md
@@ -1,0 +1,367 @@
+# API Reference
+
+This document provides a complete API reference for RAG-Anything.
+
+## Installation
+
+```bash
+pip install raganything
+```
+
+## Core Classes
+
+### RAGAnything
+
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks for contributing an API reference. Good documentation is very valuable for this project.

We cannot merge this PR in its current form because several documented APIs do not match the current codebase. For example, the document describes methods such as `ingest(...)`, `ingest_batch(...)`, and `Parser().parse(...)`, but those are not the public APIs exposed by the current implementation. There are also examples and callback snippets that do not line up with the existing interfaces, plus a few formatting issues such as trailing whitespace.

An API reference needs to be especially strict, because users will copy these examples directly. If you would like to continue with this, please regenerate the reference from the actual current public API and verify each code example against the repository before resubmitting.

Thanks again for working on the documentation.

PR #288, opened 2026-05-23
docs: add quickstart section to README

Kartalops:


LarFii:
Thanks for the contribution and for trying to make the onboarding path easier.

We cannot merge this PR as-is because the added quickstart duplicates the existing Quick Start section and, more importantly, it documents commands and APIs that do not currently exist in this project, including `python -m raganything --input ... --output ...`, `RAGAnything().process(...)`, and `result.markdown`. A quickstart is often the first thing new users try, so inaccurate examples would create a poor first experience and increase support burden.

If you would like to continue with this, please revise the quickstart around the current public APIs and examples that are already supported by the codebase, and avoid duplicating the existing Quick Start content. We would be happy to review a corrected, code-verified version.

PR #289, opened 2026-05-25
Add audio and video processing

Kartalops:
## Summary
- Add audio processing functionality
- Add video processing component

## Commits
- `35f5031` add audio process
- `2b3c83c` Merge branch 'main' into audio
- `c003cd8` add videoprocessor

## Details
This PR adds audio and video processing capabilities to RAG-Anything. The changes include:
- Audio processing module
- Video processor component

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,16 @@ The system deploys modality-aware processing units for heterogeneous data modali
   - Provides native LaTeX format support for seamless integration with academic workflows.
   - Establishes conceptual mappings between mathematical equations and domain-specific knowledge bases.
 
+- **🔊 Audio Content Analyzer**:
+  - Processes audio content including speech, music, and sound effects with specialized models.
+  - Generates comprehensive audio descriptions with transcription analysis and quality assessment.
+  - Extracts audio entities and relationships for multimodal knowledge graph integration.
+
+- **🎬 Video Content Analyzer**:
+  - Analyzes video content including visual scenes, actions, movements, and narrative structure.
+  - Processes audio-visual synchronization and temporal relationships within video sequences.
+  - Supports caption/subtitle integration for enhanced video understanding and entity extraction.
+
 - **🔧 Extensible Modality Handler**:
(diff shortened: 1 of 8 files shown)


LarFii:
Thanks for contributing audio and video processing support. This is an important direction for RAG-Anything, and the use case is very relevant.

We cannot merge this PR in its current form. I found several blocking issues during review and local validation:

- The existing GitHub check for this PR is failing on lint/format.
- `git diff --check` reports trailing whitespace in the added docs and example files.
- Running `pytest -q` on the branch collects example files as tests and fails because functions such as `test_image_format_parsing(file_path: str)` require a nonexistent pytest fixture.
- The branch does not include a current `tests/` directory, so there is no reliable regression test coverage for the new audio/video behavior.
- The audio/video implementation is a large feature addition and overlaps with other open audio/video PRs, so it needs clearer test coverage and integration boundaries before we can safely choose and merge an approach.

Audio and video support is worth pursuing, but this PR needs to be rebased onto current `main`, split into smaller reviewable pieces if possible, and covered with focused tests that pass locally and in CI. A good path would be to start with audio-only support, add unit tests around detection/transcription/chunk formatting, and keep optional heavyweight dependencies lazy so normal installs are not affected.

Thanks again for the work here. The feature direction is useful; the current branch just needs a substantial cleanup before it can be merged safely.

LarFii:
Thank you for this @Arvuno 🙏 Your integration approach (config flags, processor registration, routing, prompts, docs) was exactly the wiring that the other audio/video PRs were missing.

We consolidated everything into #292: it combines the real processing engines (faster-whisper ASR + scenedetect/keyframe/VLM dual-channel) with your integration design, and is verified end-to-end (transcribe → insert → KG → retrieve). One key difference vs this PR: #292 performs the actual transcription/scene analysis locally rather than relying on externally-provided transcripts/metadata.

Closing this as superseded by #292 — much appreciated, the integration pattern carried over directly.

PR #294, opened 2026-06-07
Fix MinerU Windows unsafe paths

ye-guan-xing:
## Summary
Fixes MinerU parsing failures on Windows when paths contain non-ASCII characters, trailing spaces, or trailing dots.
- Adds a safe-path adapter layer for MinerU.
- Applies to both `parse_pdf` and `parse_image`.
- Preserves original behavior for safe paths.
- Auto-cleans temporary files/directories after processing.

## What This Fixes
Previously, MinerU received raw input/output paths directly. On Windows, paths like:
- `sample-测试 .pdf`
- `sample-测试 .png`

could break execution or output lookup due to unsafe characters/formatting.

## Implementation
Adds a transparent safe-path layer:
1.  Unsafe input → copied to temp ASCII-only path (`input_<hash>.pdf`).
2.  Unsafe output dir → redirected to temp ASCII-only dir (`mineru_<hash>`).
3.  MinerU output → copied back to the requested directory.
4.  Temp workspace → cleaned up in `finally`.

Added MinerU helpers in `raganything/parser.py`:
- `_is_mineru_unsafe_windows_path`
- `_prepare_mineru_paths`
- `_copy_mineru_output_tree`
- `_cleanup_mineru_temp_dir`

Also fixes result reading to use the actual MinerU-safe file stem after path rewriting.

## Tests
Verified locally on Windows with:
- ✅ Unsafe PDF/image paths now parse correctly.
- ✅ Safe paths remain unchanged (no side effects).
- ✅ Temp files are properly cleaned up.
- ✅ Regression: existing MinerU flows still work on non-Windows systems.

### DIFF
--- a/raganything/parser.py
+++ b/raganything/parser.py
@@ -711,6 +711,80 @@ class MineruParser(Parser):
         super().__init__()
 
     @classmethod
+    def _is_mineru_unsafe_windows_path(cls, path: Union[str, Path]) -> bool:
+        if not _IS_WINDOWS:
+            return False
+
+        path = Path(path)
+        path_text = str(path)
+        try:
+            path_text.encode("ascii")
+        except UnicodeEncodeError:
+            return True
+
+        return any(
+            part.endswith((" ", ".")) for part in path.parts
+        ) or path.stem.endswith((" ", "."))
+
+    @classmethod
+    def _mineru_safe_path_hash(cls, path: Union[str, Path]) -> str:
+        path_text = str(Path(path).resolve())
+        return hashlib.md5(path_text.encode("utf-8")).hexdigest()[:10]
+
+    @classmethod
+    def _prepare_mineru_paths(
+        cls,
+        input_path: Union[str, Path],
+        output_dir: Union[str, Path],
+        hash_path: Optional[Union[str, Path]] = None,
+    ) -> Tuple[Path, Path, str, Optional[Path]]:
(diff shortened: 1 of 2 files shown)


LarFii (approved):
Thanks @ye-guan-xing — I verified this locally and the logic is correct.

**What I ran (on this branch):**
- Full suite `pytest tests/` → **271 passed**, including the 3 new Windows safe-path tests.
- A real-filesystem end-to-end check: stubbing *only* the `mineru` subprocess to emit a realistic output tree, then exercising the real `_prepare_mineru_paths` → `_copy_mineru_output_tree` → `_read_output_files` → cleanup across image/PDF × safe/unsafe filenames. Confirmed the temp workspace is created and then cleaned up, MinerU receives an ASCII-safe input path (no trailing space/dot), and results are copied back into the requested output dir. ✅
- `ruff check` → clean.

**One fix I applied:** `ruff format --check` was flagging `tests/testparser_kwargs.py` (the pre-3.10 `with A, B:` style; the repo's pinned `ruff-format` hook rewrites it to the parenthesized form). CI never ran on this branch, so it slipped through — I pushed a small fixup commit applying `ruff format`.

LGTM. 👍

ye-guan-xing:
Thanks for verifying and fixing the formatting issue. I'll keep an eye on CI.

PR #301, opened 2026-06-25
feat: add opt-in TwelveLabs video modality (Pegasus + Marengo)

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## Description

This PR adds an **opt-in `video` modality** to RAG-Anything, backed by the [TwelveLabs](https://twelvelabs.io) platform. It lets RAG-Anything ingest and retrieve video the same way it already handles images, tables and equations.

- **Pegasus** (`analyze`) generates a detailed transcript/description of a video. That text flows through the **existing** knowledge-graph + chunk pipeline via `BaseModalProcessor._create_entity_and_chunk` — no new retrieval path, no special-casing downstream.
- **Marengo** (`embed`) produces a 512-dim multimodal embedding, returned on the item's `entity_info` (`tl_video_embedding`) for semantic video retrieval. A companion `embed_text()` lets you score a text query in the same 512-dim space.

## Why it helps this project

RAG-Anything is explicitly multimodal, but video isn't yet a first-class modality. Pegasus + Marengo give a turnkey way to bring video understanding and video↔text semantic search into the existing graph/RAG pipeline, with no extra infra on the user's side.

## Changes Made

- `raganything/twelvelabs.py`: new `TwelveLabsModalProcessor` (mirrors `GenericModalProcessor`). Accepts `video_url`, `video_path` (uploaded as a TwelveLabs asset), or `video_id`.
- `config.py`: `enable_video_processing` flag (env `ENABLE_VIDEO_PROCESSING`), **default False**.
- `raganything.py`: registers the `video` processor only when enabled **and** `twelvelabs` is installed and a key is present; otherwise it logs a warning and skips — initialization never breaks.
- `utils.py`: dispatch routes `"video"` to the processor (falls back to `generic` when unregistered, never `KeyError`).
- `__init__.py`: optional `TwelveLabsModalProcessor` export (same try/except pattern as other optional features).
- `pyproject.toml` / `setup.py`: new `[video]` extra (`twelvelabs>=1.2.8`), also folded into `[all]`.
- `env.example`, `requirements.txt`, `README.md`: docs + config entries.

## Opt-in / non-breaking

Disabled by default. With `ENABLE_VIDEO_PROCESSING` unset, there are **zero** behavioural changes and no new hard dependency (`twelvelabs` only installs via the `[video]`/`[all]` extra).

## How it was tested

`tests/test_twelvelabs_integration.py`:
- No-network unit tests: config default, dispatch routing/fallback, and Pegasus/Marengo request wiring against a mocked TwelveLabs client.
- A live smoke test (gated on `TWELVELABS_API_KEY`, skipped in CI without it) asserting a real Marengo text embedding is **512-dim** — verified locally (`9 passed`).

Ran the repo's `ruff format --check` and `ruff check --ignore=E402` on all changed files — clean.

> Note: full Pegasus/Marengo *video* runs are server-side and can be slow (Marengo video embedding is an async, polled task). The video paths are wiring-verified against the SDK contract; the synchronous Marengo text-embedding path is verified end-to-end against the live API.

## Checklist

- [x] Changes tested locally
- [x] Documentation updated
- [x] Unit tests added

## Additional Notes

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,7 @@ pip install raganything
 pip install 'raganything[all]'              # All optional features
 pip install 'raganything[image]'            # Image format conversion (BMP, TIFF, GIF, WebP)
 pip install 'raganything[text]'             # Text file processing (TXT, MD)
+pip install 'raganything[video]'            # Video understanding & embeddings (TwelveLabs)
 pip install 'raganything[image,text]'       # Multiple features
 ```
 
@@ -578,6 +579,41 @@ async def process_multimodal_content():
 if __name__ == "__main__":
     asyncio.run(process_multimodal_content())
 ```
+
+##### Video Modality (TwelveLabs) — optional
+
+RAG-Anything ships an **opt-in** `video` modality powered by [TwelveLabs](https://twelvelabs.io):
+
+- **Pegasus** (`analyze`) generates a rich transcript/description of the video, which flows through the same knowledge-graph + chunk pipeline as every other modality.
+- **Marengo** (`embed`) produces a 512-dim multimodal embedding returned on the video item's `entity_info` (key `tl_video_embedding`) for semantic retrieval; pair it with `embed_text()` to score a text query in the same space.
+
+Enable it with the `video` extra and an API key (free tier at https://twelvelabs.io):
+
+```bash
+pip install 'raganything[video]'
+export ENABLE_VIDEO_PROCESSING=true
+export TWELVELABS_API_KEY=your-twelvelabs-api-key
+```
+
+When enabled, video items are routed automatically during document processing. You can also use the processor directly:
+
+```python
+from raganything import TwelveLabsModalProcessor
+
+video_processor = TwelveLabsModalProcessor(lightrag=rag)  # reads TWELVELABS_API_KEY
+
+description, entity_info = await video_processor.process_multimodal_content(
+    modal_content={
+        "type": "video",
+        "video_url": "https://example.com/clip.mp4",  # or video_path / video_id
+        "video_caption": ["Product demo recording"],
+    },
+    content_type="video",
+    file_path="product_demo.mp4",
+)
+```
+
+This modality is disabled by default; existing behaviour is unchanged when `ENABLE_VIDEO_PROCESSING` is unset.
 
 #### 3. Batch Processing
 
--- a/env.example
+++ b/env.example
@@ -50,6 +50,12 @@ OLLAMA_EMULATING_MODEL_TAG=latest
 # ENABLE_TABLE_PROCESSING=true
 # ENABLE_EQUATION_PROCESSING=true
 
+### Video Processing (TwelveLabs) — opt-in, requires `pip install raganything[video]`
+# ENABLE_VIDEO_PROCESSING=false
+# TWELVELABS_API_KEY=your-twelvelabs-api-key   # free tier at https://twelvelabs.io
+# TWELVELABS_PEGASUS_MODEL=pegasus1.5          # video understanding / description
+# TWELVELABS_MARENGO_MODEL=marengo3.0          # multimodal embeddings (512-dim)
+
 ### Batch Processing Configuration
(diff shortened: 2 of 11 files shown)


LarFii:
Hi @mohit-twelvelabs — thanks for this, and for the clearly opt-in, lazy-import design. A couple of things need to be resolved before it can land, one of which is timing on our side.

## 1. Namespace collision with in-progress native audio/video

We're currently building a **native, local** video+audio modality (keyframe VLM + `faster-whisper`, windowed long-media chunking) on the `feat/audio-video-modal` branch. This PR was diffed against `main`, so it doesn't see that work — but the two collide directly on every `video` knob:

- `self.modal_processors["video"]` — both register a processor under the same key
- the `enable_video_processing` config field / `ENABLE_VIDEO_PROCESSING` env var — both gate on the same flag
- the `[video]` extra in `pyproject.toml`/`setup.py` — native means `scenedetect`/`moviepy`/`faster-whisper`/`opencv`, this PR redefines it as `twelvelabs`
- the `"video"` `entity_type` and the `utils.py` dispatch branch

So as written it isn't additive — it would fork the meaning of "video." To coexist, TwelveLabs would need to be an **explicitly-named provider** rather than seizing `video`, e.g. `ENABLE_TWELVELABS_VIDEO` + a `twelvelabs` extra + a distinct processor key (or a backend-selectable `video` where `twelvelabs` is one choice). Rebasing onto `feat/audio-video-modal` and renaming the surface accordingly would remove the conflict.

## 2. Crash bug in the error fallback (independent of the above)

In `raganything/twelvelabs.py`, `process_multimodal_content`'s `except` path returns a **2-tuple**:

```python
return str(modal_content), fallback_entity
```

but the ingestion caller unpacks a **3-tuple** `(enhanced_caption, entity_info, chunk_results)`. Any exception on the video path therefore raises `ValueError: not enough values to unpack (expected 3, got 2)` and takes down the whole document ingestion instead of degrading gracefully. The native `VideoModalProcessor` returns `return str(modal_content), fallback_entity, []` here — this path should do the same, and it's worth a small test that unpacks the fallback as three values.

## Minor
- The `requirements.txt` change is only a comment, so there's **no** accidental hard dependency — the real dep is correctly confined to the `[video]` extra. 👍
- `embed_video`/`embed_text` reach into SDK internals and hardcode 512-dim; worth pinning the `twelvelabs` minor line and guarding shape drift.

Whether a paid cloud modality belongs alongside the native effort is ultimately a maintainer call, but the fallback-arity fix and the de-collision are prerequisites regardless. Happy to look again after a rebase + rename.

LarFii:
Thanks for building this, Mohit, and apologies for the long wait — this deserved an answer much sooner than it got one.

I'm closing it. Two reasons, both structural rather than about code quality:

**1. It collides with in-flight work I own.** PR #292 adds audio and video as local modalities and defines the same `enable_video_processing` / `ENABLE_VIDEO_PROCESSING` config and the same `modal_processors["video"]` slot. Merging both as written means one silently overwrites the other, so they cannot coexist in their current shapes.

**2. RAG-Anything has been merged upstream into LightRAG**, and this repository is now oriented toward algorithm validation rather than productized integrations. A vendor-hosted SaaS backend is the kind of surface that fits the LightRAG side rather than here.

One concrete bug for your records regardless of where this lands: `twelvelabs.py` mutates the global `PROMPTS` dict at import time with three English-only keys (`VIDEO_ANALYSIS_SYSTEM`, `video_analysis_prompt`, `video_chunk`). That breaks `tests/test_prompt_language.py::test_chinese_prompts_have_all_keys`, which asserts the Chinese prompt set covers every English key — reproducible on this PR against main. Any future version should either add matching `prompts_zh` entries or keep its prompts out of the shared registry.

The dependency handling here was done right — a proper `[video]` extra rather than a hard requirement — and that was noted.

PR #314, opened 2026-07-17
Validate image files before base64 encoding

adity982:
## Summary
- reuse image validation before reading files for base64 encoding
- return an empty string for non-image paths instead of encoding arbitrary files
- add regression tests for accepted and rejected paths

## Validation
- python -m py_compile raganything/utils.py tests/test_utils.py
- full pytest not run locally because lightrag is not installed in this Python 3.14 environment

### DIFF
--- a/raganything/utils.py
+++ b/raganything/utils.py
@@ -237,6 +237,8 @@ def encode_image_to_base64(image_path: str) -> str:
         str: Base64 encoded string, empty string if encoding fails
     """
     try:
+        if not validate_image_file(image_path):
+            return ""
         with open(image_path, "rb") as image_file:
             encoded_string = base64.b64encode(image_file.read()).decode("utf-8")
         return encoded_string
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,17 @@
+import base64
+
+from raganything.utils import encode_image_to_base64
+
+
+def test_encode_image_to_base64_rejects_non_image_extension(tmp_path):
+    path = tmp_path / "payload.txt"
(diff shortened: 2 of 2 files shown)


LarFii:
Thanks @adity982 — validating before encoding is a sensible hardening. Two things before this is ready:

**1. It's still marked as a draft** — let us know when you'd like it reviewed for merge.

**2. Behavior change worth confirming.** `encode_image_to_base64` currently encodes whatever path it's given; with this change it returns `""` for anything `validate_image_file()` rejects. That helper enforces (a) a fixed image-extension allowlist, (b) a **50 MB** size cap, and (c) symlink blocking. So after this change, a previously-working image that is >50 MB, reached via a symlink, or lacks a standard extension will now silently encode to an empty string, and callers can't tell the difference between "encoding failed" and "file rejected."

Could you confirm that's the intended contract? A couple of options that would make it safer:
- Log at `warning` when validation rejects a file (so it isn't a silent drop), and/or
- Have callers that already validate upstream skip the double-check.

The added tests are fine (they exercise the reject-by-extension and accept paths). Once it's out of draft and the silent-empty-string behavior is confirmed/logged, this is good to go.

adity982:
Confirmed: the validation contract is intentional as a security boundary. Paths rejected for a missing/nonstandard image extension, symlink use, or the 50 MB limit should continue to return the encoder's established failure sentinel (`""`). The rejection is not silent: `validate_image_file()` logs a warning for each of those validation failures before returning `False` (and logs an error for validation exceptions). With that clarified, this is ready for review.

PR #318, opened 2026-07-26
feat(processor): add opt-in force_multimodal_reprocess flag (fixes #154)

### LINKED ISSUE #154: [Feature Request]: Enable Cache Data Insertion for New Graph Database(or other database)

Currently, when switching to a new graph database (e.g., from Neo4j to another provider), the system fails to insert cached data into the new database. This occurs because:

The `_process_multimodal_content`  in method `Processors.py` skips the insertion step if it detects cached data ( doc_id exists in cache).
There is no mechanism to re-insertion of cached data into the new database.

### Additional Context

```python
   async def _process_multimodal_content(...):
       ...
   
       try:
            existing_doc_status = await self.lightrag.doc_status.get_by_id(doc_id)
            if existing_doc_status: 
                # Check if multimodal content is already processed
                multimodal_processed = existing_doc_status.get(
                    "multimodal_processed", False
                )

                if multimodal_processed:
                    self.logger.info(
                        f"Document {doc_id} multimodal content is already processed"
                    )
                    return
```

OliverZ-dot:
## Summary
- Adds an opt-in `force_multimodal_reprocess: bool = False` parameter to `process_document_complete()` and `insert_content_list()` that lets callers explicitly re-run image/table/equation processing for a document already marked as fully processed.
- Fixes #154: after switching to a new graph/vector storage backend, the new backend never receives the multimodal entities/relations that were only ever written to the old one, since processing is skipped once `multimodal_processed` is true.
- Default behavior is fully preserved (flag defaults to `False`); this is purely additive.
- Also hardens `chunks_list`/`chunks_count` bookkeeping to de-duplicate chunk ids, since force-reprocessing produces the same deterministic chunk ids and would otherwise duplicate entries in the list.

## Test plan
- [x] Added `tests/test_force_multimodal_reprocess.py` covering: default behavior unchanged, force flag bypasses the skip check, flag is forwarded from both public entry points, chunk id de-duplication.
- [x] Full existing test suite passes locally (265 passed, 1 skipped, skip pre-existing and unrelated).

### DIFF
--- a/README.md
+++ b/README.md
@@ -995,6 +995,39 @@ This method is particularly useful when:
 - You need to insert content from multiple sources into a single knowledge base
 - You have cached parsing results that you want to reuse
 
+#### 8. Re-processing Multimodal Content After a Storage Backend Change
+
+By default, `process_document_complete` and `insert_content_list` skip multimodal
+(image/table/equation) processing for a document that is already marked as fully
+processed, to avoid redundant LLM calls. If you switch to a **new** graph/vector
+storage backend (e.g. migrating from the default file-based storage to Neo4j),
+the new backend will not yet contain the multimodal entities/relations that were
+written to the old one — see [#154](https://github.com/HKUDS/RAG-Anything/issues/154).
+
+Pass `force_multimodal_reprocess=True` to explicitly re-run multimodal processing
+and re-populate the current storage backend for an already-processed document:
+
+```python
+# Re-run multimodal processing so the current storage backend has the
+# image/table/equation entities and relations (default is False, i.e. no change
+# to existing behavior for documents that have not switched backends)
+await rag.process_document_complete(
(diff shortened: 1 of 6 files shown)


LarFii:
Resolved the conflict with main directly on this branch (merge commit, your history untouched): the chunks_list dedup this PR carried inline landed on main via #335 with stronger semantics — chunks_count increments by newly added ids rather than being redefined as the list length, and a missing doc_status record is surfaced instead of silently skipped — so the resolution takes main's helper everywhere and keeps your force_multimodal_reprocess plumbing exactly as written. Full suite passes (345) and the flag was exercised end-to-end against a live pipeline (re-run produced no duplicate chunk ids). Merging once CI is green — thanks for the fix, and sorry it had to wait through two rounds of conflicts.

PR #226, opened 2026-03-21
fix: use valid CID font names for Chinese text rendering (fixes #24)

Exploreunive:
Fixes #24

## What

Fixes Chinese text rendering failure in PDF output caused by invalid CID font names. Chinese characters would display as black squares or fallback to default Latin fonts on Windows and macOS.

## Root Cause

`reportlab.pdfbase.cidfonts.UnicodeCIDFont` only supports a fixed set of CID font names: `STSong-Light`, `MSung-Light`, `HeiseiMin-W3`, `HeiseiKakuGo-W5`, `HYSMyeongJo-Medium`.

The old code tried to register system font names (`SimSun`, `SimHei`, `Microsoft YaHei`, `STHeiti`) which are NOT valid CID font names. All registration calls silently failed due to the `except Exception: continue` blocks, leaving no Chinese-capable font selected.

## Fix

- Replace the platform-specific invalid font name loops with `STSong-Light`, the standard cross-platform CID font for Simplified Chinese (reportlab ships the required CID resources internally)
- Only apply the fallback when WenQuanYi (Linux TrueType font) is not already registered
- Remove dead code paths for non-existent CID font names

## How to Test

```bash
python -m pytest tests/test_chinese_cid_font.py -v
```

3 tests:
- `test_stsong_light_is_valid_cid_font` — STSong-Light registers successfully
- `test_invalid_cid_font_names_raise` — old font names (SimSun etc.) correctly raise errors
- `test_all_valid_cid_cjk_font_names` — all valid CJK CID fonts are usable

Issue #24, opened 2025-07-01
[Bug]: Chinese font setup in `MineruParser.parse_text_file` never works


Glinte:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

...because you are trying to initialize `UnicodeCIDFont`s will font names that are not supported, but no one ever noticed this is failed because you suppressed all the errors. You used `["SimSun", "SimHei", "Microsoft YaHei"]` on Windows and `["STSong-Light", "STHeiti"]` on MacOS, but the only supported fonts to be used in `UnicodeCIDFont` is one of the 6 hardcoded values. By a quick read of the `UnicodeCIDFont` class docstring, I believe `CIDFont` should be used instead.

https://github.com/HKUDS/RAG-Anything/blob/9c725265857da7bba7d94b3eacfe74d089b8de98/raganything/mineru_parser.py#L647-L681

https://github.com/eduardocereto/reportlab/blob/98758940eeae30db80bbc9c555e42b8c89b86be8/src/reportlab/pdfbase/cidfonts.py#L390-L395

```python
class UnicodeCIDFont(CIDFont):
    def __init__(self, face, isVertical=False, isHalfWidth=False):
        #pass
        try:
            lang, defaultEncoding = defaultUnicodeEncodings[face]
        except KeyError:
            raise KeyError("don't know anything about CID font %s" % face)
```

https://github.com/eduardocereto/reportlab/blob/98758940eeae30db80bbc9c555e42b8c89b86be8/src/reportlab/pdfbase/_cidfontdata.py#L130-L141

```python
defaultUnicodeEncodings = {
    #we ddefine a default Unicode encoding for each face name;
    #this should be the most commonly used horizontal unicode encoding;
    #also define a 3-letter language code.
    'HeiseiMin-W3': ('jpn','UniJIS-UCS2-H'),
    'HeiseiKakuGo-W5': ('jpn','UniJIS-UCS2-H'),
    'STSong-Light': ('chs', 'UniGB-UCS2-H'),
    'MSung-Light': ('cht', 'UniGB-UCS2-H'),
    #'MHei-Medium': ('cht', 'UniGB-UCS2-H'),
    'HYSMyeongJo-Medium': ('kor', 'UniKS-UCS2-H'),
    'HYGothic-Medium': ('kor','UniKS-UCS2-H'),
    }
```

[...]

PR #333, opened 2026-08-11
feat(parser): parse .txt/.md directly instead of converting to PDF and OCR-parsing (fixes #331)

### LINKED ISSUE #331: [Feature Request]: parse .txt/.md directly instead of converting to PDF and OCR-parsing with MinerU

`parse_document()` routes `.txt`/`.md` through ReportLab (text→PDF) and then MinerU (PDF→blocks). For content that is already plain text this round-trip:

- is ~3× slower end to end (measured 52 s → 18.9 s for the same small text file when bypassing the round-trip and building the content list directly);
- is the source of a whole family of fidelity bugs — #24/#226 (CJK font rendering in the generated PDF), #316 (markdown text must be escaped for ReportLab) — none of which can exist if the text never becomes a PDF;
- burns MinerU/OCR compute on content that needs neither.

### Additional Context

Proposed behavior: for `.txt`/`.md`, build the content list directly from the text (paragraph-split on blank lines, `{"type": "text", "text": …, "page_idx": 0}`) and feed it to the existing `insert_content_list` path. We run this bypass in production glue and it produces identical downstream chunks/entities, just faster and without the PDF-rendering failure modes.

qianheng-aws:
## Description

`.txt`/`.md` files were rendered to PDF with ReportLab and then re-parsed with the OCR pipeline. For content that is already text this round-trip is ~3× slower end to end (measured 52 s → 18.9 s for the same small text file), burns OCR compute, and is the source of a family of rendering-fidelity bugs (#24, #226, #316) that cannot exist if the text never becomes a PDF.

## Related Issues

Fixes #331. Related: #24, #226, #316 (all bugs in the PDF-rendering leg this removes from the text path).

## Changes Made

- `Parser.parse_text_file` (new, on the base class): reads the file (UTF-8 with BOM handling, falling back to gbk/latin-1/cp1252) and splits it into MinerU-style content blocks. Blank lines delimit paragraphs; in markdown, ATX headings become their own blocks carrying `text_level`, matching how MinerU marks headings in its content lists.
- `MineruParser.parse_text_file` and `PaddleOCRParser.parse_text_file` overrides (which did `convert_text_to_pdf` → `parse_pdf`) are removed; both inherit the direct path. `convert_text_to_pdf` remains available for direct callers.
- The encoding-fallback read is factored into `_read_text_with_fallback`, shared with `convert_text_to_pdf`.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- Tests: `tests/test_plain_text_direct.py` (13 cases: paragraph/heading splitting, CRLF handling, BOM stripping, gbk fallback, `parse_document` routing for both parsers with the PDF path monkeypatched to fail, extension/missing-file errors). Full suite: 248 passed.
- Lint: `ruff==0.6.4 check --ignore=E402` and `ruff format --diff` clean on changed files (matching the pre-commit config).
- Behavioral note: text files no longer produce parser output artifacts on disk (there is no conversion), and blocks carry `page_idx: 0` since plain text has no pages. Callers consuming the returned content list are unaffected.

### DIFF
--- a/raganything/parser.py
+++ b/raganything/parser.py
@@ -34,6 +34,7 @@ import subprocess
 import tempfile
 import threading
 import logging
+import re
 import time
 import urllib.parse
 import urllib.request
@@ -410,25 +411,7 @@ class Parser:
                 raise ValueError(f"Unsupported text format: {text_path.suffix}")
 
             # Read the text content
-            try:
-                with open(text_path, "r", encoding="utf-8") as f:
-                    text_content = f.read()
-            except UnicodeDecodeError:
-                # Try with different encodings
-                for encoding in ["gbk", "latin-1", "cp1252"]:
-                    try:
-                        with open(text_path, "r", encoding=encoding) as f:
-                            text_content = f.read()
-                        cls.logger.info(
-                            f"Successfully read file with {encoding} encoding"
-                        )
-                        break
-                    except UnicodeDecodeError:
-                        continue
-                else:
-                    raise RuntimeError(
-                        f"Could not decode text file {text_path.name} with any supported encoding"
-                    )
+            text_content = cls._read_text_with_fallback(text_path)
 
             # Prepare output directory
             if output_dir:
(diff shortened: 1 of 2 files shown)


LarFii:
Merged — thanks for the careful work here and on the sibling PRs.

One design note for the record (from a question that came up in review): the old PDF round trip was originally intended to carry markdown-referenced images into the pipeline, but it never actually did — the ReportLab leg renders `alt` as literal text and embeds zero images (verified empirically: 0 embedded images in the generated PDF). So this PR loses nothing on that front. Real markdown image extraction — emitting an image block with the path resolved relative to the source file — is now a small follow-up on top of your `_text_to_content_blocks`, and is planned.

Issue #328, opened 2026-08-11
[Bug]: process_document_complete_lightrag_api() pre-registers doc_status and is silently skipped by LightRAG's dedup — same mechanism as #277, not covered by #278


qianheng-aws:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

#277 reported that pre-registering the target doc_id in `doc_status` before handing content to LightRAG makes `apipeline_enqueue_documents`' key-existence check (`filter_keys`) treat the document as a historical duplicate and abort text processing. #278 fixed this for `insert_content_list` by deferring the doc_status write.

The sibling method `process_document_complete_lightrag_api()` still has the same pattern on current `main`: it upserts the content-based `doc_id` with `DocStatus.HANDLING` (Step 1, `processor.py` ~line 2014) **before** calling `insert_text_content_with_multimodal_content(..., ids=doc_id)`. LightRAG's dedup then sees the id as already enqueued, logs `Duplicate document detected`, and skips indexing entirely.

The result is a silent no-op: the document parses (MinerU extracts blocks), status reaches `processed`, but **zero chunks and zero entities are written** and queries return nothing. There is no error anywhere — the failure is only visible as an absence.

Two secondary problems in the same method:

- It writes fields into `doc_status` that LightRAG's `DocProcessingStatus` dataclass does not declare (`content`, `multimodal_content`, `scheme_name`), so LightRAG Server's own WebUI fails to parse the record: `Error parsing doc doc-…: unexpected keyword argument 'content'`. This is the same family as #73 (which hit `multimodal_processed`).
- It seeds `"created_at": ""` (`processor.py:1880`, `:1937`). On storage backends that map `created_at` as a date (e.g. OpenSearch/Elasticsearch), the upsert is rejected with `mapper_parsing_exception: cannot parse empty date` — and because LightRAG's DocStatus upsert bulks with `raise_on_error=False`, the write is dropped **silently**; the immediately following `get_by_id()` returns `None` and the method dies with `TypeError: 'NoneType' object is not a mapping`. Invisible on the default file-based storage, which is presumably why it hasn't been reported.

### Steps to reproduce

1. Use `process_document_complete_lightrag_api()` against a LightRAG instance (observed with lightrag-hku 1.4.x-series and OpenSearch-backed storages; the pre-registration pattern is unchanged on current main).
[...]

PR #336, opened 2026-08-12
feat(processor): opt-in page provenance for text chunks via LightRAG's sidecar pipeline (fixes #330)

### LINKED ISSUE #330: [Bug]: separate_content() concatenates text blocks before chunking, so page_idx never reaches text chunks

### Describe the bug

MinerU records a `page_idx` on every content block, but `separate_content` (`raganything/utils.py`, `text_content = "\n\n".join(text_parts)`) concatenates all text blocks into one string **before** LightRAG token-chunks it. The page index does not survive into the resulting chunks.

Measured on a mixed corpus: all 128 multimodal chunks carried a `page_idx` (each multimodal block becomes exactly one chunk, so it survives), while all 57 plain-text chunks carried none.

### Why it matters

Any downstream feature that needs page provenance for text — page-anchored citations, figure-to-chunk association by page range, PDF viewers jumping to the source page — is unbuildable for exactly the chunks that carry the bulk of the document. LightRAG itself has been moving toward positional metadata (HKUDS/LightRAG#3081 emits page-level positions from `page_idx` in its own MinerU path), but content ingested through RAG-Anything's `separate_content` loses the information before LightRAG ever sees it.

### Expected Behavior

Preserve per-block `page_idx` through chunking — e.g. chunk per page-run instead of one concatenated string, or carry a block-offset→page map alongside the text so chunks can be annotated after splitting.

qianheng-aws:
## Description

`separate_content` joins every text block into one string before LightRAG chunks it, so MinerU's per-block `page_idx` never reaches text chunks (#330) — measured on a mixed corpus: all 128 multimodal chunks carried a `page_idx`, all 57 text chunks carried none. Page-anchored citations, reader jump-to-page and page-based figure association are unbuildable for exactly the chunks that carry the bulk of a document.

Rather than inventing a new provenance mechanism, this PR routes RAG-Anything's own MinerU parse through the provenance pipeline **LightRAG already ships**: engine adapters emit an IR whose blocks carry positions (page + bbox), `write_sidecar` persists it, and the paragraph-semantic chunker records each chunk's source blocks as `sidecar.refs` — joinable back to page numbers and bounding boxes. RAG-Anything documents currently bypass all of it because they are inserted as one raw string.

```
content_list (text blocks only)
  → MinerUIRBuilder.normalize_from_workdir   (LightRAG's own adapter)
  → write_sidecar                            (LightRAG's own writer)
  → apipeline_enqueue_documents(process_options="P")
  → full_docs row marked parse_format="lightrag" + sidecar_location
  → pipeline dispatches by format to ReuseParser
  → P chunker consumes blocks.jsonl → chunks carry sidecar.refs
```

Verified end to end on a real 61-page slide deck before wiring: **16/16 text chunks carried block refs resolving to page positions** (multi-block chunks correctly resolve to page ranges, e.g. `pages=[2,3,4,5,6]`).

## Related Issues

Fixes #330. Aligns with the direction of HKUDS/LightRAG#3081 (page-level positions in LightRAG's own MinerU path).

## Changes Made

- New `raganything/sidecar_ingest.py`: `sidecar_available()` (feature detection), `text_only_content_list()` (mirrors `separate_content`'s split — multimodal items stay with the modal processors exactly as before; their chunks already carry `page_idx` one-to-one), and `insert_text_with_sidecar()` (the pipeline above, with `ainsert`-equivalent duplicate-skip semantics).
- `process_document_complete` Step 3 routes through it when opted in; any failure falls back to plain insertion.
- New config flag `preserve_page_provenance` (`PRESERVE_PAGE_PROVENANCE`, **default off**) — opt-in because the chunking strategy changes from fixed-token to paragraph-semantic. Also bypassed when a custom `split_by_character` is requested (the P chunker owns splitting) or the installed lightrag predates the sidecar subsystem.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- Tests: `tests/test_sidecar_provenance.py` (8 cases): the text-only filter; real `MinerUIRBuilder`+`write_sidecar` against a recording fake LightRAG asserting the enqueue/merge contract and that sidecar blocks carry page positions (skipped gracefully when the installed lightrag lacks the sidecar subsystem); duplicate-skip without forging state; processor wiring (opt-in routes, failure/unavailable/custom-split/flag-off all fall back to plain insertion).
- Two contracts are mutation-tested, each learned from a failed integration run: the full_docs update must **merge** (enqueue stores `process_options` there; replacing the row silently reverts the document to the fixed-token chunker), and enqueue must select `"P"` (the default chunker ignores the sidecar).
- Full suite: 243 passed.
- Lint: `ruff==0.6.4 check --ignore=E402` and `ruff format` clean on changed files.

### DIFF
--- a/raganything/config.py
+++ b/raganything/config.py
@@ -51,6 +51,16 @@ class RAGAnythingConfig:
     )
     """Enable equation content processing."""
 
+    preserve_page_provenance: bool = field(
+        default=get_env_value("PRESERVE_PAGE_PROVENANCE", False, bool)
+    )
+    """Insert text through LightRAG's sidecar pipeline so text chunks keep
+    block-level page provenance (``sidecar.refs`` joinable to page numbers
+    and bounding boxes — see issue #330). Requires a lightrag-hku build with
+    the sidecar subsystem and changes how text is chunked (paragraph-semantic
+    instead of fixed-token); falls back to plain insertion when unavailable
+    or when a custom ``split_by_character`` is requested."""
+
     # Batch Processing Configuration
     # ---
     max_concurrent_files: int = field(
--- a/raganything/processor.py
+++ b/raganything/processor.py
@@ -1749,14 +1749,54 @@ class ProcessorMixin:
                         doc_id=doc_id,
                     )
                 insert_start = time.time()
-                await insert_text_content(
-                    self.lightrag,
-                    input=text_content,
-                    file_paths=file_name,
-                    split_by_character=split_by_character,
-                    split_by_character_only=split_by_character_only,
-                    ids=doc_id,
-                )
+                inserted_via_sidecar = False
+                # Page-provenance path (#330): route the text through
+                # LightRAG's sidecar pipeline so chunks keep block-level page
+                # positions. Only when opted in, no custom split is requested
+                # (the P chunker owns splitting), and the installed lightrag
+                # ships the sidecar subsystem; any failure falls back to the
+                # plain insertion below.
+                if (
+                    getattr(self.config, "preserve_page_provenance", False)
+                    and split_by_character is None
+                ):
+                    from raganything.sidecar_ingest import (
+                        insert_text_with_sidecar,
+                        sidecar_available,
+                    )
+
+                    if sidecar_available():
+                        try:
+                            await insert_text_with_sidecar(
+                                self.lightrag,
+                                content_list=content_list,
+                                doc_id=doc_id,
+                                file_name=file_name,
+                                document_name=Path(file_path).name,
+                                sidecar_parent_dir=output_dir,
+                            )
+                            inserted_via_sidecar = True
+                        except Exception as sidecar_exc:
+                            self.logger.warning(
+                                "Sidecar insertion failed for "
+                                f"{file_path} ({sidecar_exc}); falling back "
+                                "to plain text insertion"
+                            )
+                    else:
+                        self.logger.info(
+                            "preserve_page_provenance is set but the "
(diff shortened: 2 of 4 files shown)


LarFii:
Thanks for this — the design is right, and routing through LightRAG's own sidecar pipeline rather than inventing a parallel provenance mechanism was the correct instinct.

I'm closing it for a reason you had no way to know: **RAG-Anything has been merged upstream into LightRAG.** For LightRAG 1.5+, this functionality lives in LightRAG itself, and changes here sync there. That makes `lightrag-hku<1.5` a permanent pin for this repo rather than a temporary migration state.

The practical consequence for this PR: `raganything/sidecar_ingest.py` imports a subsystem that first exists in lightrag 1.5.0, and 1.4.15's `apipeline_enqueue_documents` has no `process_options` parameter. Merging it would ship the module as dead code behind `sidecar_available()`, plus a documented user-facing knob (`preserve_page_provenance` / `PRESERVE_PAGE_PROVENANCE`) that can only ever log "lacks the sidecar pipeline" — and its two substantive tests skip in CI for the same reason.

#330 still stands as a real issue on the pinned line. If you want to take another run at it, the shape that works without 1.5 is: keep the single concatenated string that `separate_content` produces today, carry a block-offset → `page_idx` map alongside it, and annotate chunks with their page after LightRAG has split them — one `doc_id`, one `doc_status` row, no change to the ingest contract. That last part matters: #344 attempted #330 by splitting ingestion into `doc_id#p{i}` ids and broke the canonical document row, re-opening what #328/#334 fixed.

Sorry for the wait on this one.

PR #344, opened 2026-08-22
fix: preserve page boundaries through text chunking (#330)

### LINKED ISSUE #330: [Bug]: separate_content() concatenates text blocks before chunking, so page_idx never reaches text chunks

lmhche66-coder:
`separate_content()` previously joined every text block into a single string before LightRAG token-chunks it, so the per-block `page_idx` that MinerU records never survived into text chunks — only multimodal chunks kept page provenance. This breaks page-anchored features (citations, figure-to-chunk association, source-page navigation) for the bulk of a document.

This change groups text by `page_idx` into page-runs and inserts each run separately so page boundaries survive chunking: `separate_content` now returns `(text_content, page_texts, multimodal_items)`, and the three insertion paths pass `page_texts` with page-encoded ids (`doc_id#p{i}`). Consecutive short pages are merged into a run (above `MIN_PAGE_RUN_CHARS`) to avoid tiny, low-context chunks; a run never mixes non-adjacent pages.

Verification: 5 new unit tests (12/12 pass); end-to-end with local Ollama + lightrag 1.4.16 produced page-homogeneous chunks with zero cross-page chunks.

Closes #330

### DIFF
--- a/raganything/processor.py
+++ b/raganything/processor.py
@@ -1703,7 +1703,7 @@ class ProcessorMixin:
                 doc_id = content_based_doc_id
 
             # Step 2: Separate text and multimodal content
-            text_content, multimodal_items = separate_content(content_list)
+            text_content, page_texts, multimodal_items = separate_content(content_list)
 
             # LightRAG creates the initial doc_status entry during text insertion.
             # Pre-registering the same doc_id here makes LightRAG treat a fresh
@@ -1739,11 +1739,11 @@ class ProcessorMixin:
                 insert_start = time.time()
                 await insert_text_content(
                     self.lightrag,
-                    input=text_content,
-                    file_paths=file_name,
+                    input=page_texts,
+                    file_paths=[file_name] * len(page_texts),
                     split_by_character=split_by_character,
                     split_by_character_only=split_by_character_only,
-                    ids=doc_id,
+                    ids=[f"{doc_id}#p{i}" for i in range(len(page_texts))],
                 )
                 await self._upsert_doc_status(
                     doc_id,
@@ -2010,7 +2010,7 @@ class ProcessorMixin:
                 doc_id = content_based_doc_id
 
             # Step 2: Separate text and multimodal content
-            text_content, multimodal_items = separate_content(content_list)
+            text_content, page_texts, multimodal_items = separate_content(content_list)
 
             # LightRAG creates the initial doc_status entry during text
             # insertion. Pre-registering the same doc_id here makes LightRAG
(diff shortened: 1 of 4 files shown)


LarFii:
Thanks for taking a run at #330 — it's a real gap and the page-run grouping idea is a reasonable instinct.

I'm closing it, because as written it regresses a fix that landed recently. Merged onto current main, this branch fails four existing tests:

```
FAILED tests/test_insert_content_list.py::test_insert_content_list_defers_status_until_after_text_insert
FAILED tests/test_insert_content_list.py::test_process_document_complete_defers_status_until_after_text_insert
FAILED tests/test_lightrag_api_doc_status.py::test_text_content_is_indexed_not_skipped_as_duplicate
FAILED tests/test_lightrag_api_doc_status.py::test_real_doc_id_is_not_registered_before_ainsert
```

The last two are the regression suite from #334, which fixed a bug where a document reached `processed` status with zero chunks and zero entities written because LightRAG's dedup treated the fresh insert as a historical duplicate. Those tests going red is the substantive problem, not a fixture mismatch: inserting under `f"{doc_id}#p{i}"` ids means LightRAG never creates a row for the bare `doc_id`, so the canonical document row becomes an empty shell (`content_length=0`, `chunks_list=[]`, no `full_docs` entry) — which is the same class of breakage #328/#334 were merged to fix. `separate_content()` also changed to a 3-tuple while callers still unpack two, which is where the `TypeError: unhashable type: 'list'` at `tests/test_insert_content_list.py:61` comes from.

Separately, and this would matter even if the above were fixed: `#p{i}` is a **run ordinal**, not a page index. Consecutive short pages are merged into one run, so a figure-only page shifts every subsequent ordinal, and nothing in the repo ever parses the id back into a page number. So it doesn't deliver what #330 actually asks for.

If you want to try again, the shape that works without changing the ingest contract: keep the single concatenated string `separate_content` produces today, carry a block-offset → `page_idx` map alongside it, and annotate chunks with their page **after** LightRAG has split them — one `doc_id`, one `doc_status` row.

PR #345, opened 2026-08-22
feat: parse .txt/.md directly instead of the PDF round trip (#331)

### LINKED ISSUE #331: [Feature Request]: parse .txt/.md directly instead of converting to PDF and OCR-parsing with MinerU

lmhche66-coder:
`.txt` / `.md` files are currently routed through `convert_text_to_pdf` (ReportLab) and then OCR-parsed by MinerU — a round trip that is ~3× slower (52s → 18.9s), introduces fidelity bugs (CJK font rendering #24/#226, markdown escaping #316), and burns OCR compute on content that needs none.

This change parses plain text / markdown directly: `parse_document` now detects `.txt` / `.md` and builds the content list straight from the file (paragraph-split on blank lines, `{"type": "text", "text": …, "page_idx": 0}`), feeding the existing insert path unchanged. UTF-8 reading with invalid bytes replaced keeps CJK and other non-ASCII text intact.

Verification: 6 new unit tests (13/13 pass); a Chinese markdown file parses into the expected text blocks with `page_idx=0`.

Closes #331

### DIFF
--- a/raganything/processor.py
+++ b/raganything/processor.py
@@ -19,6 +19,7 @@ from raganything.utils import (
     separate_content,
     insert_text_content,
     insert_text_content_with_multimodal_content,
+    build_content_list_from_text_file,
     get_processor_for_type,
     format_table_body,
     get_equation_text_and_format,
@@ -507,6 +508,15 @@ class ProcessorMixin:
                         output_dir=output_dir,
                         **kwargs,
                     )
+            elif ext in [".txt", ".md"]:
+                self.logger.info(
+                    "Detected plain text/markdown file, building content list "
+                    "directly (bypassing the text->PDF->OCR round trip)..."
+                )
+                content_list = await asyncio.to_thread(
+                    build_content_list_from_text_file,
+                    file_path,
+                )
             elif ext in [
                 ".doc",
                 ".docx",
--- a/raganything/utils.py
+++ b/raganything/utils.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import base64
 import inspect
+import re
 from typing import Dict, List, Any, Tuple
 from pathlib import Path
 from lightrag.utils import logger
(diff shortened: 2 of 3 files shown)


LarFii:
Thanks for this — and to be clear up front, there's nothing wrong with it. It's passing cleanly against main, and the encoding work is good.

I'm closing it because #333 arrived first for the same issue (#331) and wins the comparison on where the change belongs. #345 adds the `.txt`/`.md` branch one layer up, in `ProcessorMixin`, and never touches `parser.py`. The consequence is that `MineruParser`, `PaddleOCRParser` and `BatchParser` still take the ReportLab → OCR round trip, so the same `.md` file behaves two different ways within one release depending on which entry point you came through. #333 removes the `parse_text_file` overrides at the parser layer instead, so every parser inherits the direct path and there's one behavior.

I checked one axis where I expected #345 to come out ahead and it didn't: `DoclingParser().parse_document('x.md')` raises the same `ValueError: Unsupported file format: .md` under main, #333 and #345 alike. #333 is actually the only one of the three that gets Docling closer to working, since it inherits a functioning `parse_text_file` from the base class.

Genuinely sorry to close a working PR over timing and layering. If you want to keep contributing here, two things came out of reviewing both PRs that are unclaimed and would be welcome:

1. **Markdown images.** Neither the old PDF path nor #333 extracts `alt` — I verified the ReportLab leg never imported PIL's `Image` and rendered the markdown source as literal text, producing zero embedded images. On top of #333's block builder this is a small, self-contained change: emit an `{"type": "image", "img_path": ...}` block with the path resolved relative to the source file.
2. **Fenced code blocks.** `#` inside ```-fenced blocks is currently treated as a heading. An `in_fence` toggle fixes it.

PR #347, opened 2026-08-24
feat(parser): support native MinerU content list v2

### LINKED ISSUE #224: [Feature Request]:

RAG-Anything already advertises MinerU integration and direct content list insertion, and there has also been recent compatibility work for MinerU 2.x (for example, the hybrid backend output directory issue and the merged MinerU-2 field-name fix).

However, from practical inspection of current MinerU outputs, a ZIP bundle may now contain both:

- legacy flat `*_content_list.json`
- newer `content_list_v2.json`

In my case, the current pipeline still ends up consuming the legacy flat `*_content_list.json`, while `content_list_v2.json` appears to be a richer and more structured intermediate representation.

Why this matters

The legacy flat content list contains many low-semantic structural blocks such as:

- `page_number`
- `header`
- `list`
- `footer`

When these are passed into the current RAG-Anything ingestion path, they can amplify noise in downstream semantic extraction and multimodal graph construction.

By contrast, `content_list_v2.json` appears to provide a more semantically organized representation (for example, grouped blocks and higher-level block types such as `title`, `paragraph`, `image`, etc.), which seems more suitable as the upstream document representation for RAG ingestion.

Observed behavior

In MinerU output bundles, I can see files like:

- `*_content_list.json`
- `content_list_v2.json`
- `full.md`
- `layout.json`

But the current ingestion path appears to still prefer the legacy `*_content_list.json`.

Request

Please consider adding **native support for `content_list_v2.json`** in RAG-Anything, including:

1. Detect and prefer `content_list_v2.json` when available
2. Add a native parser/adapter for the v2 schema
3. Clarify whether v2 is officially supported, experimental, or unsupported
4. Document the expected MinerU output contract(s) for current RAG-Anything versions

Suggested acceptance criteria

- RAG-Anything can ingest MinerU bundles where `content_list_v2.json` is present
- The ingestion path does not require falling back to legacy flat `*_content_list.json`
- Core semantic content types from v2 are properly routed into text / multimodal processing
- Structural-only blocks (such as page numbers or similar layout artifacts) are not accidentally amplified as semantic entities

### Additional Context

_No response_

smf-h:
## Summary

- Add a standalone adapter for MinerU 3.x `content_list_v2.json` (`list[list[dict]]`) and normalize it into RAG-Anything's existing flat content-list contract.
- Route titles, paragraphs, lists/indexes, inline/interline equations, images, tables, charts, code, and algorithms into the existing text/multimodal pipeline while preserving `page_idx`, `bbox`, `anchor`, heading levels, captions, footnotes, and visual metadata.
- Filter `page_header`, `page_footer`, `page_number`, `page_aside_text`, and `page_footnote` by default; expose `include_layout_blocks=True` for explicit opt-in.
- Add deterministic MinerU artifact discovery for exact and generic v2/legacy filenames. Document-scoped output trees win over shared roots; within one bundle v2 wins over legacy, with exact stem names preferred within the same schema.
- Keep legacy ingestion resilient: malformed, structurally invalid, or semantic-empty v2 output falls back to an available legacy content list.
- Include `include_layout_blocks` in parser-cache isolation and document the supported contract in English and Chinese READMEs.

Algorithms are represented as the existing compatible `type="code", sub_type="algorithm"` shape, while retaining the original MinerU v2 type metadata. Unknown future v2 block types are skipped with a warning so one schema extension does not discard supported content.

## Compatibility

Legacy `*_content_list.json` and `content_list.json` bundles remain supported. Existing relative-media path resolution and traversal checks are applied after v2 normalization as well.

## Validation

- `pytest -q`
- `pytest -q tests/test_mineru_content.py`
- `ruff check raganything/mineru_content.py tests/test_mineru_content.py`
- `ruff format --check raganything/mineru_content.py tests/test_mineru_content.py`
- `ruff check raganything/parser.py --select E9,F63,F7,F82`
- `python -m compileall -q raganything`

Closes #224

### DIFF
--- a/README.md
+++ b/README.md
@@ -1153,6 +1153,7 @@ await rag.process_document_complete(
     backend="pipeline",          # Parsing backend: pipeline|hybrid-auto-engine|hybrid-http-client|vlm-auto-engine|vlm-http-client.
     source="huggingface",        # Model source: "huggingface", "modelscope", "local"
     # vlm_url="http://127.0.0.1:3000" # Service address when using backend=vlm-http-client
+    include_layout_blocks=False,  # Keep MinerU v2 headers, footers, and page numbers only when needed
 
     # Standard RAGAnything parameters
     display_stats=True,          # Display content statistics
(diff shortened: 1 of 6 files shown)


LarFii:
Pushed one small commit on top: `_source_path` now rejects a media source whose basename is empty ("images/", ".", "..") — such a path resolved to a directory downstream and surfaced as a misleading per-item "Error generating image description" (IsADirectoryError swallowed by the try-block). Everything else is untouched.

For the record, two review notes I considered and deliberately did NOT apply: the `candidate_key` ordering stays yours — within a shared bundle a generic v2 almost certainly belongs to the same document (RAG-Anything's own flow gives each file a unique output dir via `_unique_output_dir`), so preferring the richer format is right, and your test pins it. Solid PR — verified the adapter end-to-end against a live pipeline as well.

PR #349, opened 2026-08-30
fix(example): tolerate LM Studio model discovery failures

huyua9:
## Description

The LM Studio integration example treated `/models` discovery as a hard connection requirement. When an OpenAI-compatible LM Studio server does not expose a working models endpoint, the example returned early and never ran its chat-completion health check.

Model discovery is now best-effort; chat completion remains the authoritative health check. The client cleanup path is also safe when initialization fails.

## Changes Made

- Continue to `test_chat_completion()` when `client.models.list()` fails.
- Add a regression test with a mocked unavailable `/models` endpoint.

## Validation

- `python3 -m py_compile examples/lmstudio_integration_example.py tests/test_lmstudio_integration_example.py`
- `pytest -q tests/test_lmstudio_integration_example.py` (1 passed)
- `ruff check examples/lmstudio_integration_example.py tests/test_lmstudio_integration_example.py`
- `git diff --check`

### DIFF
--- a/examples/lmstudio_integration_example.py
+++ b/examples/lmstudio_integration_example.py
@@ -28,13 +28,13 @@ from typing import List, Dict, Optional
 from dotenv import load_dotenv
 from openai import AsyncOpenAI
 
-# Load environment variables
-load_dotenv()
-
 # RAG-Anything imports
 from raganything import RAGAnything, RAGAnythingConfig
 from lightrag.utils import EmbeddingFunc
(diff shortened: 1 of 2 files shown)


LarFii:
Pushed a one-line format commit on top (pinned ruff-format v0.6.4, what CI runs, wanted the `spec_from_file_location` call wrapped). No functional changes — the model-discovery fallback itself checks out.

PR #359, opened 2026-09-12
fix(parser): decode local Markdown image paths

lllleolin-max:
## Description

Markdown image destinations containing URL escapes are treated as literal filenames. With a local `figure 1.png`, parsing `Figure` currently produces a text block, so the image never reaches multimodal processing.

Decode local destinations once before resolving the filesystem path. This also handles UTF-8 names, escaped `#` and `%`, and absolute paths. Literal `+` stays unchanged, and a filename containing the literal text `%20` can be referenced as `%2520`. Image titles, remote URLs, and unresolved references retain their existing behavior.

## Related Issues

No linked issue; reproduced through `MineruParser.parse_text_file` with local Markdown and image files.

## Changes Made

- Use the existing `urllib.parse` import to decode local image destinations.
- Add regression coverage through the actual text-file parser for encoded and ordinary paths, titles, absolute paths, and unresolved references.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed — independent Codex agent review
- [x] Documentation updated (if necessary) — inline comment; no public API change

## Additional Notes

- Same focused test file: **7 failed / 17 passed** before the fix; **24 passed** after it.
- Full suite: `python -m pytest tests -q` — **406 passed, 1 skipped**.
- Changed files pass Ruff 0.6.4 formatting/lint checks and `git diff --check`.
- Tested on Windows with Python 3.12.14 and LightRAG 1.4.16. No OCR models or live LLM services were used.

AI assistance: implemented and validated with Codex, with independent Codex agent review. No independent human review is claimed.


LarFii:
Pushed one commit on top (plus a merge of current main so CI runs against it). The decode is right, but on its own it regressed a case that resolved in v1.4.1: a file literally named `a%2Bb.png` referenced verbatim became a text block, because only the decoded form (`a+b.png`) was checked. Reproduced side by side — main: `a%2Bb.png` → image, `figure%201.png` → text; this PR before the push: exactly inverted.

The added commit tries the decoded target first and falls back to the verbatim one when they differ, and adds two verbatim-percent rows to your resolver table. Your `%2520` double-encoding path still works as documented. Merging once CI is green — thanks for the careful test table.

